PR #146, opened 2024-09-13
Update image domains and refactor localization files

guillermoscript:
This pull request includes several changes. First, the `next.config.js` file is updated to include 'wp.villademarcos.com' in the list of allowed image domains. Second, the localization files are refactored to include components. Third, the localization for components is added. Fourth, the localization files are refactored to include components and update the image domain in the LoginLayout. Finally, the `.env.example` file is refactored to add the WP contact form ID. These changes improve the image handling and localization in the codebase.


angel-afonso (commented):
LGTM 🚀

PR #330, opened 2026-06-14
refactor(subscriptions): make plan/subscription system provider-agnostic

guillermoscript:
Closes #280. Implements the reviewer-approved first slice: provider subscription fields migration, Stripe webhook lookup fix, and plan creation refactored onto the existing `products.ts` provider pattern.

## What changed

**1. Migration** `20260615120000_provider_agnostic_subscriptions.sql`
- `plans.stripe_product_id` → `provider_product_id`, `stripe_price_id` → `provider_price_id` (RENAME, preserves existing data; idempotent via `DO`/`information_schema` guard).
- `subscriptions` gains `payment_provider TEXT NOT NULL DEFAULT 'manual'` + `provider_subscription_id TEXT`.
- Partial index on `provider_subscription_id` for webhook lookups.

Verified against the live schema before writing: `plans` had `stripe_*` + `payment_provider`; `subscriptions` had none of the three columns — confirming the webhook's `stripe_subscription_id` lookup was matching a **non-existent column** (silent no-op).

**2. `IPaymentProvider`** (`lib/payments/types.ts`) — optional `createSubscription` / `cancelSubscription` / `getSubscription` + `CreateSubscriptionParams` / `ProviderSubscription`.

**3. `ManualPaymentProvider`** — no-op subscription methods (manual billing is DB-managed).

**4. `StripePaymentProvider`** — real subscription lifecycle (`subscriptions.create/cancel/update`), with a `mapSubscription` helper. `current_period_end` reads via `as any` (Stripe SDK v2025 type-shape change — matches the existing pattern in `webhook/route.ts`).

**5. `app/actions/admin/plans.ts`** — removed the module-level `new Stripe(...)`; all create/update/archive paths now use `getPaymentProvider()` with `type: 'subscription'` recurring prices, mirroring `products.ts`. **Creating a `manual` plan now needs no Stripe credentials.**

**6. `app/api/stripe/webhook/route.ts`** — `customer.subscription.deleted` + `invoice.payment_failed` now look up by `(provider_subscription_id, payment_provider='stripe')`.

**7. `app/actions/admin/subscriptions.ts`** — provider-aware cancellation (calls the provider when `provider_subscription_id` is set; manual no-ops; provider errors are logged and don't strand the DB update). Dropped the unused `getStripe` import.

## Verification
[...]

Issue #280, opened 2026-05-29
refactor(subscriptions): make plan/subscription system provider-agnostic (currently Stripe-hardcoded)


guillermoscript:
## Summary

The **product** payment system is properly abstracted behind an `IPaymentProvider` interface and supports Stripe, PayPal, and manual payments interchangeably. The **subscription/plan** system was built separately and is hardcoded to Stripe throughout — at the action layer, the DB schema, and the webhook layer. This blocks adding any other recurring billing provider without a full rewrite.

---

## User Story

> As a school admin, I want to offer my students recurring subscription plans paid through different payment methods (Stripe, PayPal, bank transfer, or any future provider), so that students in my region can subscribe using whatever payment option works for them — not just the ones Stripe supports.

---

## Business Implications

- **LATAM market**: Stripe has limited availability and high card decline rates in several LATAM countries. Schools need alternatives (MercadoPago, local bank transfers, PayPal). Right now subscriptions are Stripe-only at the code level.
- **Manual subscriptions work today** — if a student pays via bank transfer and the admin confirms, a subscription is created with no Stripe involvement. But the plan *creation* still requires Stripe credentials, which is inconsistent.
- **Webhook lifecycle is broken today**: the Stripe webhook for `customer.subscription.deleted` queries a column (`stripe_subscription_id`) that does not exist on the `subscriptions` table. Stripe recurring billing lifecycle events (renewals, cancellations, payment failures) are silently failing.
- **Vendor lock-in risk**: adding a secondary processor requires a full rewrite rather than adding a new provider implementation.

---

### What already works
- `IPaymentProvider` interface exists at `lib/payments/types.ts`
- `stripe-provider.ts`, `manual-provider.ts`, `paypal-provider.ts` all implemented
- `plans` table already has a `payment_provider` column
- Manual plan payments via `payment-requests.ts` are fully provider-agnostic today

### What is broken / hardcoded

**1. `app/actions/admin/plans.ts` — bypasses the abstraction entirely**

```ts
import Stripe from 'stripe'
const stripe = new Stripe(process.env.STRIPE_SECRET_KEY!, { apiVersion: '2026-02-25.clover' })
// createPlan, updatePlan, archivePlan all call stripe.* directly
// Compare: products.ts uses getPaymentProvider(formData.paymentProvider) correctly
```

**2. `plans` table has Stripe-specific column names**

| Current | Should be |
|---|---|
| `stripe_product_id` | `provider_product_id` |
[...]

PR #334, opened 2026-06-15
feat(payments): Lemon Squeezy + Solana providers — one-time split & native auto-pull subs (#280)

guillermoscript:
## Provider-agnostic payments — Lemon Squeezy + Solana (one-time split + native auto-pull subs)

Phases 4–6 of issue #280, on top of the merged provider contract + native-Stripe work (#330/#331). Lets a school pick a **payment provider per plan/product** and keeps the subscription model working across all of them — driven by LATAM needs where Stripe is weak. Adds **three** providers behind the existing `IPaymentProvider` contract + capability flags, with one unified checkout entry, one unified webhook entry, and one shared billing-event dispatcher.

> **Status:** money flows are verified at the DB/trigger layer, in LiteSVM, and through the checkout UI wiring; the full unit suite is green (37/37) and CI is passing. A **live charge has not been run** — no provider credentials are provisioned and the on-chain legs need a funded wallet (see *Not yet proven*). Migrations are **local-only, not pushed to cloud** (app has no prod users yet).
>
> **Reachability caveat:** ~~the **student checkout half** of each journey is wired end-to-end, but the **school-admin setup half is not** — there is no UI yet to select these providers or enter their config, so today they're enabled only by writing DB rows directly (see *User-journey readiness*).~~ **No longer accurate — see the 2026-07-03 update below.** The admin self-serve config UI (`ba0bbdc3`) shipped after this was written; both halves of the journey are now reachable through the app.

---

### Update — 2026-07-03

This description was written when the branch was 14 commits / 37 files; it's now **24 commits / 85 files**, so treat the sections below as a design reference, not a current changelog. Since the original post:

- **Admin self-serve provider UI shipped** (`ba0bbdc3`) — Settings → Payment toggles + wallet form, plan/product provider dropdowns now include Lemon Squeezy/Solana/`solana_subs` with conditional fields. The *Reachability caveat* and *User-journey readiness* table below are stale as a result — admin config is no longer the missing half.
- **Lemon Squeezy one-time orders completed** (`ed215f00`) — `order_created`/`order_refunded` are now mapped; one-time LS purchases previously took payment without enrolling the buyer.
- **In-app Phantom checkout + money-viewing dashboards added** (`7f1ad7cd`, `30064cf2`) — desktop wallet flow (no QR scan needed), student billing hub, admin payouts/invoices, super-admin revenue dashboard.
- **USD-priced Solana settlement** (`3bab27c5`) — USDC default (1:1, no oracle) with an opt-in native-SOL path priced off a live Pyth oracle and locked at checkout.
- **Manual QA pass across all three personas on devnet** (PR comment, 2026-06-21) — full checkout→on-chain→enrollment→billing-view path verified with a real funded wallet. Logged 4 findings, **not yet fixed**: (1) the Solana Pay tx-request URL used a single global `NEXT_PUBLIC_APP_URL`, breaking checkout for any tenant other than the pinned host — **fixed today, see below**; (2) an abandoned Solana checkout leaves a `pending` transaction blocking retry with a generic error; (3) the admin Transactions table shows "N/A" for the Solana payment method column; (4) the revenue dashboard showed $0 platform fee on a Solana sale that should have had one (needs a clean re-check, test-setup confound).
- **Merged master in** (this update) — the PR was `CONFLICTING` (a one-line `.gitignore` comment collision); resolved, now `MERGEABLE`.
- **Applied and verified the 6 previously-unapplied local migrations** (`20260615150000`–`20260617130000`) against local Supabase; `scripts/test-payments-integration.ts` passes end-to-end against the current schema.
- **Fixed finding (1) above** — Solana checkout now derives its tx-request origin from the incoming request (matching the pattern already used in `/api/auth/callback`) instead of the single global env var, so every tenant subdomain gets a correctly-scoped QR link. Findings (2)–(4) are UX/display issues, not merge-blockers — recommend tracking as a fast-follow.
- Build/typecheck/lint/unit-tests all still pass on the current HEAD (`npx tsc --noEmit` clean, `npm run build` exit 0, 52/52 unit tests).

---

## How much changed

**37 files · +9,458 / −1,703 · 14 commits.**

| Area | Files | What |
|---|---|---|
| `lib/payments/` | 9 | The 3 providers, fee-split, native-subs engine, shared dispatcher, types |
| `app/api/payments/` | 4 | Unified checkout + unified webhook + Solana tx/verify routes |
| `app/api/cron/` | 2 | `solana-pull` crank (new) + capability-aware `expire-subscriptions` |
| `supabase/migrations/` | 4 | Provider columns, `tenant_payment_wallets`, native-subs tables, signature-consume index |
| `tests/unit/` | 4 | 37 assertions (split math, LS webhook HMAC, dispatcher branching, pull-resume) |
| UI / i18n | 3 | `checkout-form.tsx` redirect+QR branching, en/es strings |

**Commit groups:**
- **Feature** (`86c361cb`, `36bca189`, `de44a7e5`) — LS provider, Solana fee-split, native auto-pull subs.
- **Correctness/tests** (`e949370b`, `e58974fb`, `798d4eec`, `6cd7b9ee`, `c4e7a525`, `80dae5d3`) — DB integration test, checkout feature-list parse fix, cron fail-closed, idempotent flip, unit tests, dependency spike.
- **Security review** (`2ae6136d`, `f9dadaee`, `43500b68`) — H1/H2/M1/M2/M4 (see *Security hardening* below).
- **Infra** (`7fd6ae16`) — lockfile regenerated to match CI's npm 10.

---

## Providers added

| Provider | Buys | Flow the student sees | How we confirm the money |
|---|---|---|---|
| `lemonsqueezy` | one-time **or** subscription | Card, on Lemon Squeezy's hosted page (redirect away + back) | HMAC `X-Signature` webhook → shared `dispatchBillingEvent` |
| `solana` | one-time product | Wallet QR, one tap, atomic 2-leg split (school + platform fee) | On-chain: `/solana/verify` runs `verifySplitTransfer` (both legs) |
| `solana_subs` | subscription | Wallet QR → on-chain SUBSCRIBE (delegated auto-pull) | On-chain SUBSCRIBE confirm + first pull; a cron crank renews each period |

Stripe (Connect) and `manual` (offline payment-request) keep their existing dedicated paths untouched — the unified checkout route explicitly rejects them so there is exactly **one** creation path per provider.

---

## How it works — user journeys

Everything starts at the same checkout card (`components/public/checkout-form.tsx`). The plan/product row carries a `payment_provider`, and the form branches on it:
- **`lemonsqueezy`** → "redirect provider": leaves the app to a hosted page.
- **`solana` / `solana_subs`** → "QR provider": shows a Solana Pay QR and polls for on-chain confirmation.
- **free** → instant enroll; **`manual`** → offline payment-request tab; **`stripe`** → existing inline flow.

The single shared primitive is a **pending `transactions` row** created up front. Its `transaction_id` is our correlation id — it round-trips back to us on the Lemon Squeezy webhook (as `custom_data.reference`) or to the Solana verify endpoint (as the on-chain reference pubkey). When that row flips to `successful`, the existing `after_transaction_update` trigger creates the subscription + entitlements (or enrolls the product) — **the same trigger every provider already used**, so downstream access logic is unchanged.

### A · Lemon Squeezy (card, Merchant of Record)

LS remits tax itself, so there's **no Connect split** — the school is paid out by LS, the platform reconciles separately (a deferred follow-up).

**One-time product or first subscription payment:**
1. Student clicks **Pay** → `POST /api/payments/checkout` inserts the pending tx and calls `LemonSqueezyProvider.createCheckoutSession`. We embed `reference = transaction_id` (plus `userId`/`tenantId`) in LS `checkout_data.custom`.
2. Response is `{kind:'redirect', url}`; the browser navigates to LS's hosted checkout. Student pays on LS's domain.
3. LS fires a signed webhook → `POST /api/payments/webhook/lemonsqueezy`:
   - `verifyWebhook` checks the `X-Signature` HMAC-SHA256 (timing-safe) on the **raw** body.
   - The event is persisted to `webhook_events` keyed by a synthesized-but-unique `providerEventId` (`event:subId:updated_at`) → **idempotent**, so a redelivery or two renewals are never collapsed.
   - `normalizeWebhookEvent` maps `subscription_created` → `subscription.activated` (carrying the echoed `userId/tenantId`), then `dispatchBillingEvent` **binds the flip to the echoed owner** and flips our pending tx → `successful`. Trigger creates the subscription + entitlements.
4. Student is redirected back to `/dashboard/student?checkout=success`.

**Recurring lifecycle (all webhook-driven, no cron):**
- `subscription_payment_success` → `subscription.renewed` → `extend_subscription_period` pushes `end_date` to LS's `renews_at` (idempotent — period comes from the event, not `now()`).
- `subscription_payment_failed` → `subscription.past_due` (log-only; access continues until period end).
- `subscription_cancelled` / `subscription_expired` → status set; access ends at period end (LS only cancels at period end).
- Refunds: `refund()` hits LS's subscription-invoice refund API.

Because LS emits renewal webhooks, the `expire-subscriptions` cron **skips** LS subs (capability `emitsRenewalWebhooks`) — they're never prematurely expired.

### B · Solana one-time (`solana`) — atomic fee split

Solana Pay TRANSFER allows only one recipient, so the split is a Solana Pay **transaction request**: our server builds one transaction with **two** transfer legs (school share + platform fee) that the wallet signs once and the runtime executes atomically.

1. Student clicks **Pay** → `POST /api/payments/checkout` inserts the pending tx, generates a unique **on-chain reference pubkey**, stores it on the tx, returns `{kind:'qr', url}` where `url` is a Solana Pay tx-request link.
2. The form renders the QR and **polls `POST /api/payments/solana/verify` every 3s**.
3. Student scans with a Solana wallet. The wallet fetches the link → `/api/payments/solana/tx` returns the serialized 2-leg split tx (school = remainder, platform = floor of `revenue_splits.platform_percentage`; the reference is attached to the school leg). Student approves once; both legs settle atomically.
4. `/solana/verify` runs `verifySplitTransfer`: `findReference` locates the tx, then it asserts **both legs** by exact `(destination, amount)` — for USDC the destination is the recipient's **ATA**, which binds the mint. Wrong amount/recipient → it throws and we never confirm.
5. On a valid match we flip the tx → `successful` **and consume the on-chain signature** (`provider_charge_id`, unique index) so one on-chain payment can back only one order (H1). Trigger enrolls the product. Poll sees `confirmed:true` → toast + redirect.

No webhook/secret is needed — the only way to make verify succeed is a real on-chain transfer of the exact split to our recipients carrying our unique reference. The `status='pending'` guard makes the flip idempotent against repeated polls.

### C · Solana native subscription (`solana_subs`) — on-chain auto-pull

There is no card-on-file on Solana, so recurring billing is done with an **on-chain Subscriptions program**: the student authorizes a delegated pull with a per-period cap; we (the whitelisted "puller"/merchant) pull each period. The on-chain program enforces the cap and the period gate, so an early or over-cap pull simply reverts.

**Subscribe + first charge:**
1. Student clicks **Pay** on a `solana_subs` plan → checkout returns a Solana Pay QR whose link is `/api/payments/solana/subscribe-tx`. The student's wallet pubkey (subscriber) is captured **server-side** at that step.
2. Wallet approves the **SUBSCRIBE** tx (creates the on-chain `SubscriptionDelegation` PDA for `(merchant, planId, subscriber)`).
3. Poll → `/solana/verify` (subs branch): confirms the SUBSCRIBE landed (`findReference`), confirms the delegation exists on-chain (`getSubscriptionState`), flips the tx → `successful` keyed by the delegation **PDA** (the stable renewal match key), persists the on-chain coordinates on the subscription row, then fires the **first split pull** (school leg, then platform leg).

**Each renewal (`/api/cron/solana-pull`, run hourly, `CRON_SECRET`-gated):**
For every active `solana_subs` sub whose on-chain period has rolled over, the crank submits the two-leg split pull and extends our DB period by one cycle. The on-chain cap + period gate already prevent double/unauthorized charges; an in-period or over-cap pull reverts and that row is skipped. Cancellation on-chain (delegation closed / `expiresAt` passed) → the crank expires our row → trigger revokes access.

Because these are self-cranked (not webhook-renewed), `expire-subscriptions` skips them via capability `supportsNativeSubscriptions`.

---

## User-journey readiness (what's reachable in the app today)

The journeys above describe what happens **once a plan/product row has the right `payment_provider` + config**. Here's how much of that a real user can drive through the UI right now:

| Step | Actor | Built? | Where / gap |
|---|---|---|---|
| See checkout, branch to LS redirect or Solana QR, pay, poll → enroll | Student | ✅ | `app/[locale]/(public)/checkout/page.tsx` + `components/public/checkout-form.tsx`; en/es strings present |
| Select **Lemon Squeezy / Solana / solana_subs** for a **plan** | Admin | ❌ | `components/admin/plan-form.tsx` dropdown lists only `manual` / `stripe` / `paypal` |
| Select a provider for a **product** | Admin | ❌ | no product provider selector exists |
| Enter the **Lemon Squeezy variant id** (`provider_price_id`) | Admin | ❌ | no field; checkout 400s without it |
| Enter the school's **Solana wallet** (`tenant_payment_wallets`) | Admin | ❌ | no UI; verify/pull 400 without it |
| Set the platform-fee % (`revenue_splits`) | Admin | ➖ | reuses the existing Stripe-Connect split model |

**Net:** the rails + student checkout are complete and tested; **the admin-facing "turn it on and configure it" UI is the missing half.** Today a provider is enabled only by seeding the DB rows directly (as `scripts/test-payments-integration.ts` and `scripts/test-lemonsqueezy-pipeline.ts` do). A small follow-up — extend the plan/product form dropdown with the three providers + conditional fields for variant id and wallet address, plus a per-tenant wallet/settings panel — makes these journeys self-serve. Recommend tracking it as its own issue before any school is expected to use this unaided.

---

## Security hardening (from the #334 review)

Reviewed with 4 parallel read-only agents + manual verification of the top findings; fixed on-branch (CI green, covered by the unit suite):

- **H1 — one payment can't back N orders.** A single on-chain tx can list multiple reference keys. `/solana/verify` now writes the confirmed **signature** to `provider_charge_id`, guarded by a partial-unique index `(payment_provider, provider_charge_id) WHERE status='successful'` → a replay against a second order hits `23505` → **409**.
- **M1 — webhook can't activate someone else's tx.** LS `subscription.activated` echoes `userId/tenantId` in `custom_data`; the dispatcher now **binds the flip to that owner** and refuses on mismatch (a signed event can't flip a pending tx by sequential id alone).
- **M2 / M4 — subscribe hardening.** Plan lookup is tenant-scoped (`.eq('tenant_id', …)`); `provider_metadata` is **write-once** (`.is(..., null)`).
- **H2 — auto-pull residual (documented + made safe, behavior deferred).** The on-chain cap + period gate already prevent double/unauthorized charge; the residual is revenue-*accuracy* only (a partial-leg RPC failure could miss the platform fee for a period; DB price drift vs the immutable on-chain cap). Shipped a **monotonic-safe `alreadyPulledBase` resume primitive** (unit-tested) but deferred wiring the on-chain value + charging from on-chain terms until a **devnet test plan** exists — see follow-ups.
- Downgraded an agent "SPL mint unchecked" CRITICAL → LOW: the verify destination is the USDC **ATA**, which the token program binds to the mint.

---

## Architecture highlights

- **Capability-driven, never by provider name.** A `PROVIDER_CAPABILITIES` map; the expiry cron gates on `emitsRenewalWebhooks || supportsNativeSubscriptions`, so push-renewal (Stripe/PayPal/LS) **and** native-pull (`solana_subs`) subs are excluded from premature cron-expiry — only self-managed-period providers (one-time Solana, manual) are cron-expired.
- **Unified checkout** `/api/payments/checkout`: insert pending tx → call provider → return `{kind: redirect|qr|client_secret}`; rolls the tx → `failed` on provider error so the partial-unique index doesn't block a retry.
- **Unified webhook** `/api/payments/webhook/[provider]`: verify → persist (idempotent on `provider_event_id`) → normalize → dispatch → 200. `manual`/`solana` are deliberately **not** exposed here (no signed webhook = no unauthenticated mutation surface; Solana confirms on-chain).
- **Shared dispatcher** `webhook-dispatch.ts` is the single source of truth for applying normalized billing events across every provider.
- **Native subs** (`solana-subscriptions.ts`) is isolated on `@solana/kit` 6.9.0 (web3.js v2) and crosses into the v1 stack only via base58 string addresses.

---

## Testing

- `npm run test:unit` → **37/37** (`solana-split`, `solana-subscription-pull`, `webhook-dispatch`, `lemonsqueezy-webhook`, smoke). `npx tsc --noEmit` clean. `npm run build` exit 0. CI green.
- `scripts/test-solana-subscriptions.ts` — 16/16 (LiteSVM, no network): plan/subscribe/pull/cap-revert/period-advance/cancel + unsigned-subscribe + ensurePlanOnChain.
- `scripts/test-lemonsqueezy-pipeline.ts` — HMAC verify (valid/tampered/missing), normalize, activation→trigger→sub+entitlements, period-align to `renews_at`, idempotency, renewal extend.
- `scripts/test-payments-integration.ts` — network-free DB money path: confirmed `solana_subs` tx → trigger creates sub + 2 entitlements; crank renewal advances period; cancel revokes; capability-aware expiry gating.
- **Migrations** apply in order on a fresh `supabase db reset`, no conflicts. **Test-isolation fix**: the pipeline tests' `delete entitlements` is now scoped to the test plan's `source_id` (previously corrupted alice's seeded entitlements).

## Not yet proven (need funded wallet / real account)

- Two-leg **on-chain per-period cap** end-to-end (LiteSVM only exercised single pulls): both pulls in one period succeed, a third reverts.
- First-charge (verify) vs crank **double-pull** guard in period 1, on a real cluster.
- Lemon Squeezy **live sandbox** (needs a real LS test account + tunnel).

## Out of scope / follow-ups

- **The durable Solana fix — `plans/012-solana-kit-migration.md`** (scoped): migrate the payment rails off legacy `@solana/web3.js` + `@solana/pay` + `@solana/spl-token` onto `@solana/kit`, clearing the 5-high/7-moderate advisory class (incl. the unpatched `bigint-buffer` DoS) no `overrides` pin can fix. **Phase 5 of that plan is the home for the deferred H2 on-chain work + its devnet harness.** Gated on this PR merging.
- Migrations not pushed to cloud (deliberate — no prod users).
- MoR (Lemon Squeezy) payout reconciliation undesigned (deferred).


guillermoscript:
### Update — in-app Phantom checkout + UX polish (`7f1ad7cd`)

Wired the **real student-facing in-app wallet flow** for Solana (the QR path couldn't cover desktop: the Phantom browser extension can't scan a QR, and a first-time subscription needs two signed txs a single scan can't do).

**Added**
- **`CheckoutForm` → "Pay with Phantom"**: detects `window.solana`, connects, runs the two-step loop (init authority → subscribe for first-time subscribers; single tx for one-time), signs sign-only, submits via the server, polls `/verify`. QR stays the one-time fallback; subscriptions without a wallet show honest "open in Phantom / install the desktop extension" guidance + a phantom.app link.
- **`/api/payments/solana/submit`** (new): authed relay that broadcasts the wallet-signed tx and confirms it before returning. Sign-only + server submit is required because Phantom's own simulate trips on these instructions, and the init tx must be confirmed before the subscribe tx is requested.

**UX polish (from a user-story audit of the admin + student flows)**
- Plan/Product create copy was Stripe-specific ("created in Stripe") for every provider → now provider-neutral; the per-method hint already adapts.
- The Solana wallet alert on the plan/product forms now links to **Settings → Payment** instead of just naming it.

i18n en + es throughout.

**Verification**
- `tsc --noEmit` → 0 errors; both message files valid JSON.
- Drove the real app (local sessions injected to bypass the flaky base-ui login): student `solana_subs` checkout renders the correct no-wallet guidance branch; admin settings + plan create/edit render with provider-aware copy and the Settings link; `POST /api/payments/solana/submit` returns 401 unauthenticated.
- Earlier this session, the full `solana_subs` path was verified live on **devnet**: subscribe → delegation → first 80/20 split pull → renewal crank → DB period-extend (`extend_subscription_period`).
- ⏳ The live two-prompt Phantom signing is the one path that needs a real wallet click-through (financial action) — not automatable.

_Note: Bug B (the earlier `0x64`/`136` subscribe revert) was resolved in `ba0bbdc3`+`b93cd1ad` — `@solana/subscriptions@^0.4.0-rc.1` + init-authority split into its own confirmed tx._

guillermoscript:
## ✅ Manual Verification — Solana payments across personas

**Verdict: PASS.** Drove the live app (local dev, devnet) end-to-end through admin, student, and super-admin. The full Solana checkout completed on-chain and produced a real enrollment. Captured 3 real findings + dev-env sharp edges below.

**Method:** `npm run dev` on `localhost:3005` against local Supabase + Solana **devnet**; in-app Phantom checkout signed by a real funded devnet wallet. Tenant = Default School (`owner@e2etest.com` admin/super-admin, `student@e2etest.com`).

### Steps (running app)
1. ✅ **Admin · config** — Settings → Payment: Solana toggle + "Also accept native SOL" render; receiving wallet saved → `Solana wallet saved` toast.
2. ✅ **Student · checkout UI** — `/checkout`: item + price, **USDC/SOL settlement toggle** with correct copy ("USDC is a 1:1 US-dollar stablecoin" / "Priced in USD; you pay the equivalent in SOL at the live exchange rate, locked when you start checkout"), **Pay with Phantom** + QR fallback.
3. ✅ **Student · pay (SOL)** — Pay with Phantom → `Connecting…` → `Approve in Phantom to sign…` → `Submitting to the network…` → `Confirming on-chain…` → **"You're all set! Your purchase was successful."**
4. ✅ **On-chain + DB** — tx `10028` = `successful`, `settlement=sol`, base `13,684,724` lamports @ `$73.07/SOL`; `provider_charge_id` = on-chain signature `5KkCv97c…Zsyfo`, **`err: None`, finalized**. Oracle math exact: `$1.00 ÷ $73.074 = 0.013685 SOL` ✓. **Enrollment** in course → `active`. Payment correctly bound 1:1 to the order.
5. ✅ **Student · billing** — Purchase History shows `VERIFY Solana Course Access · Solana · ✓ Paid · $1.00`.
6. ✅ **Admin · money view** — Transactions: **Total Revenue $1.00 / Pending $0 / Failed 0**, row `10028 Successful $1.00`. Revenue dashboard: Total $1.00, **Revenue by Product → "VERIFY Solana Course Access $1.00"**, monthly trend bar.
7. ✅ **Super-admin** — Tenant Management loads (Super Admin badge), both tenants listed.
8. 🔍 **USDC path probe** — built + signed fine; failed on-chain `InstructionError[3, Custom:1]` = SPL-Token `InsufficientFunds` (test wallet lacks the **custom devnet USDC mint** `H5cxrc…9J2k`). Env/funding, not app. Same school wallet has prior `err:null` USDC transfers, so the split program path is sound.
9. 🔍 **Abandoned-checkout probe** — a checkout left mid-flow blocks the next attempt (see Finding 2).

### Findings
- ⚠️ **(1) Solana Pay tx-request URL is a single global `NEXT_PUBLIC_APP_URL`** (`lib/payments/solana-provider.ts:106`, `solana-subscriptions-provider.ts:100`). Solana Pay does require an absolute reachable URL, but: (a) it must be a publicly-reachable host in prod, and (b) using one global value means *all tenants'* Solana tx requests route through that single host — and if it's set to a specific tenant subdomain, checkout from any other host throws `Failed to fetch`. Worth a comment + a prod-config check.
- ⚠️ **(2) Abandoned Solana checkout leaves a `pending` transaction that blocks retry.** The `transactions_unique_product` partial index covers `status IN ('pending','successful')`, so a half-finished checkout (user closed Phantom, RPC hiccup) makes the next attempt fail server-side with a generic **"Failed to create transaction"** (`duplicate key`). No auto-expiry/cancel of stale pending Solana intents, and the surfaced error doesn't explain it. Consider expiring stale `pending` solana tx (or upserting the existing pending) + a clearer message.
- ⚠️ **(3) Admin Transactions "Payment Method" column shows `N/A`** for the Solana row (provider not surfaced), while the student Billing view correctly shows "Solana". Inconsistent — admin can't tell a payment was Solana from the table.
- 🔎 **(4) Revenue dashboard "Platform Fees" = $0.00** for the Solana sale. Default School is on the **Free plan (10% fee per docs)**, so a non-zero platform fee was expected. (Caveat: my test set the school receiving wallet == platform wallet, which muddies the on-chain split, so this needs a clean re-check — flagging as "verify fee accounting for Solana in the revenue view," not a confirmed bug.)
- 🔎 **Dev-env sharp edges (not prod bugs, but rough):** (a) login **silently** fails to persist a session on any non-`lvh.me` host because the browser client pins the auth cookie to `.lvh.me` via `NEXT_PUBLIC_PLATFORM_DOMAIN` — a new dev on `localhost` gets no error and an infinite login bounce; (b) post-login client redirect drops the port (`router.push` → `localhost:80`); (c) public devnet RPC (`api.devnet.solana.com`) 429-rate-limits the submit (1 send + up to 30 status polls) — a dedicated RPC is advisable for reliable settlement/testing.

*Verified manually via Claude-in-Chrome on the running app; on-chain state confirmed against Solana devnet.*

PR #413, opened 2026-07-16
feat: refresh impeccable skill assets

guillermoscript:
## Summary\n- Refresh the impeccable skill assets and supporting metadata across the workspace\n- Add the exercise workspace component and update the exercise chat flow\n- Tighten exercise component typing so the touched files pass lint\n\n## Validation\n- npx eslint components/exercises/essay-exercise.tsx components/exercises/exercise-chat.tsx


guillermoscript:
Closing as obsolete — the skill this refreshes no longer exists in the repo.

`master` (5abc4239) drops the `impeccable` umbrella skill entirely: `.agents/skills/impeccable/` and its `.claude/skills/impeccable` symlink are gone, and the `impeccable` entry was removed from `skills-lock.json`. The per-verb skills it used to wrap (`adapt`, `animate`, `audit`, `critique`, `polish`, …) stay as independent skills sourced from `pbakaus/impeccable`.

So the 100 files here would land under a path nothing reads any more. If any individual asset in this PR is still worth keeping, it should go to the per-verb skill that owns it instead.

Issue #479, opened 2026-07-19
QA: credentialed end-to-end sandbox testing for PayPal & Binance Pay (follow-up to #466 / PR #478)


guillermoscript:
Follow-up to #466 / PR #478 (implements PayPal in full + Binance Pay). That PR shipped with code + build + unit-test verification only — **no live sandbox run**, because no PayPal sandbox or Binance Pay merchant credentials were available in the dev environment. This issue tracks the credentialed end-to-end QA that must happen after the PR merges, before either provider is enabled for a real school.

## Prerequisites (setup)

- [ ] Apply the migration (`supabase db push`) — adds `binance` to the products/plans `payment_provider` CHECK constraints (`20260719180000_add_binance_payment_provider.sql`).
- [ ] **PayPal sandbox app** (developer.paypal.com): set `PAYPAL_CLIENT_ID`, `PAYPAL_CLIENT_SECRET`, `PAYPAL_ENVIRONMENT=sandbox`.
- [ ] **PayPal webhook**: register `https://<host>/api/payments/webhook/paypal` with events `PAYMENT.CAPTURE.COMPLETED`, `PAYMENT.CAPTURE.REFUNDED`, `BILLING.SUBSCRIPTION.ACTIVATED`, `BILLING.SUBSCRIPTION.CANCELLED`, `BILLING.SUBSCRIPTION.EXPIRED`, `BILLING.SUBSCRIPTION.SUSPENDED`, `BILLING.SUBSCRIPTION.PAYMENT.FAILED`, `PAYMENT.SALE.COMPLETED`; set the webhook id as `PAYPAL_WEBHOOK_ID`. Local testing needs a public tunnel (ngrok/cloudflared) since PayPal must reach the endpoint.
- [ ] **Binance Pay merchant account**: set `BINANCE_PAY_API_KEY`, `BINANCE_PAY_API_SECRET`; configure the merchant webhook URL to `https://<host>/api/payments/webhook/binance`.
- [ ] Admin → Settings → Payment: enable PayPal and Binance Pay for the test tenant.

## PayPal — one-time (product)

- [ ] Create a paid product with provider PayPal → a Catalog Product id (`PROD-…`) is stored in `provider_product_id` and a synthetic `PAYPAL-ONETIME:…` id in `provider_price_id`.
- [ ] Student checkout → redirected to PayPal sandbox approve page → after approval, returned via `/api/payments/paypal/capture` to the success page.
[...]

PR #478, opened 2026-07-19
feat(payments): implement PayPal (full) and Binance Pay providers (#466)

guillermoscript:
## Description

Issue #466 asked for a decision on the two advertised-but-unimplemented providers: **implement or remove**. Per maintainer decision, this PR **implements both**.

### PayPal — full implementation (`lib/payments/paypal-provider.ts`)

Every method was a `console.log` stub; now a complete provider (no SDK — raw `fetch`, mirroring the Lemon Squeezy provider):

- **OAuth2** client-credentials token management with caching; sandbox/live base URL via `PAYPAL_ENVIRONMENT`.
- **Catalog** (`createsCatalog: true`): Catalog Products API + Billing Plans API. Subscription prices become Billing Plans (`P-…`); one-time prices get a deterministic synthetic id (Orders v2 charges raw amounts — PayPal has no one-time price object). Archive/restore of products are documented no-ops (PayPal's API has none); plan archive = deactivate.
- **Checkout** (`createCheckoutSession`, `kind: 'redirect'`): Orders v2 (`intent: CAPTURE`) for one-time; Billing Subscriptions API for plans. Owner-binding `reference|userId|tenantId` is packed into PayPal's single 127-char `custom_id` and unpacked on webhook normalize, satisfying the fail-closed owner guard in `webhook-dispatch.ts`.
- **New capture route** `app/api/payments/paypal/capture/route.ts`: Orders v2 doesn't auto-capture, so the approve redirect returns through this route, which captures the order, dispatches `payment.succeeded` through the shared dispatcher, and redirects to the success page. Open-redirect guarded (same-origin `next` only) and replay-safe (`ORDER_ALREADY_CAPTURED` → idempotent re-dispatch).
- **Webhooks**: `verifyWebhook` via PayPal's verify-webhook-signature API (requires `PAYPAL_WEBHOOK_ID`); `normalizeWebhookEvent` maps `PAYMENT.CAPTURE.COMPLETED/REFUNDED`, `BILLING.SUBSCRIPTION.ACTIVATED/CANCELLED/EXPIRED/SUSPENDED/PAYMENT.FAILED`, and `PAYMENT.SALE.COMPLETED` (renewals; fetches `next_billing_time` for the period extension) onto the 8 internal event types.
- **`cancelSubscription`**, **`getSubscription`**, **`refund`** (full + partial with currency lookup).
- Capabilities unchanged from what was advertised, plus `supportsPlanChange: false` (no native in-place swap — plan changes use the #463 supersession path).

[...]

Issue #466, opened 2026-07-18
Remove or implement advertised-but-unimplemented providers (PayPal, Binance)


guillermoscript:
Part of #458 (Phase 4 — provider hygiene).

## Problem
- **PayPal**: `lib/payments/paypal-provider.ts:48-143` — every method is a `// TODO` stub. Yet `PROVIDER_CAPABILITIES` (`lib/payments/types.ts:132-140`) advertises `supportsNativeSubscriptions: true, createsCatalog: true`, the unified webhook lists it as SUPPORTED (returns 501 in practice), and admin provider settings can enable it. Nothing user-facing works.
- **Binance**: in the provider type union + capability table, but `getPaymentProvider` throws "not yet implemented" (`lib/payments/index.ts:37-38`).

## Wanted
Decide per provider: **implement or remove**. If not implementing now:
- Remove/disable from `PROVIDER_CAPABILITIES` (or add an `implemented: false` flag the UI respects), the webhook `SUPPORTED` list, and the admin payment-settings wizard (PR #339 `enabledProviders` gating).
- Guard `plans`/`products` rows from being created with a non-functional `payment_provider`.

## Acceptance
- [ ] No UI surface (admin settings, plan/product creation, checkout) offers a provider that cannot complete a real payment end-to-end.
- [ ] Capability table matches reality; webhook allowlist matches implemented verify/normalize.

guillermoscript:
🤖 An AI agent (Claude Code, run by @guillermoscript) is picking this issue up. Decision per maintainer: **implement both providers** rather than remove them.

## Current behavior

- `lib/payments/paypal-provider.ts` is a placeholder: every catalog method is a `console.log` stub returning fake ids; no OAuth, no `createCheckoutSession`, no webhook verify/normalize, no subscriptions, no refunds. Yet `PROVIDER_CAPABILITIES` advertises full native-subscription + catalog support, the unified webhook route lists `paypal` as SUPPORTED (returns 501 in practice), and the admin settings wizard can enable it.
- `binance` exists only in the provider type union + capability table; `getPaymentProvider('binance')` throws "not yet implemented". The products/plans `payment_provider` CHECK constraints don't even allow `binance`.

## Approach

Implement both against the existing capability-driven abstraction (checkout `kind` dispatch + `NormalizedBillingEvent` + provider-agnostic `dispatchBillingEvent`), based on latest `origin/master` (includes the #463 `updateSubscription`/`supportsPlanChange` contract).

[...]

PR #490, opened 2026-07-23
fix(payments): categorize binance_enabled under Payment settings, not general (#479)

guillermoscript:
## What & why

Closes part of #479 (credentialed-QA follow-up to #466 / PR #478). While setting up the live PayPal sandbox portion of that QA, a survey of `app/actions/admin/settings.ts` and `components/admin/payment-settings-form.tsx` — needed to enable PayPal for a test tenant — turned up a second, credential-independent bug in the same admin Settings > Payment surface #479's "UI polish check" covers. The live PayPal/Binance sandbox runs are tracked separately; see the QA record comment on #479.

## The bug

`getAllSettingsByCategory()` hardcodes a `categoryMap` that assigns each `tenant_settings` row to a tab (`general`/`email`/`payment`/`enrollment`) for the admin Settings page. It listed `stripe_enabled`, `paypal_enabled`, `lemonsqueezy_enabled`, and `solana_enabled` under `'payment'` — but not `binance_enabled`. Any key missing from the map falls through to `'general'`.

`PaymentSettingsForm` reads `settings.binance_enabled?.value?.enabled ?? false` from the `payment` bucket to set the Binance Pay switch's initial state. Since the row never landed there, **the Binance Pay toggle always rendered unchecked on page load, regardless of the saved DB value.** An admin who enabled it, navigated away, and came back would see it off — and because the form resubmits every field from its current checkbox state, hitting Save from that stale view would silently flip `binance_enabled` back to `false` in the database.

This is a display/round-trip bug only: `getEnabledPaymentProviders()` (used by checkout, product creation, and plan creation to decide which providers to show) queries `binance_enabled` directly and was unaffected — actual provider gating elsewhere in the app kept working correctly throughout.

Verified live in a local Code Academy Pro tenant: enabling the toggle persisted `binance_enabled: {"enabled": true}` to `tenant_settings`, but reloading the Payment tab with the bug present showed the switch unchecked; with the fix, it correctly shows checked after reload.

## The fix

Add `binance_enabled: 'payment'` to the `categoryMap` in `getAllSettingsByCategory()`, and `binance_` to the equivalent prefix list in the sibling `getSettings(category)` helper (same root cause, same file — currently unused elsewhere in the app, but left consistent rather than leaving an identical latent bug behind).

## Test plan

- [x] New regression test `tests/unit/admin-settings-category-map.test.ts` — asserts a `binance_enabled` row lands under `payment`, not `general`, and stays grouped alongside the other payment toggles.
- [x] `npx vitest run tests/unit/admin-settings-category-map.test.ts tests/unit/admin-enabled-providers.test.ts tests/unit/admin-provider-actions.test.ts tests/unit/paypal-binance-providers.test.ts` → 33/33 pass.
- [x] `npx tsc --noEmit` → clean.
- [x] `npx eslint` on changed files → only the 8 pre-existing `no-explicit-any` errors already present on `app/actions/admin/settings.ts` before this change (same lines, unrelated to this diff) — no new errors introduced.
- [x] Live-verified in the browser against a local Code Academy Pro tenant: reproduced the bug (toggle shows unchecked after reload despite `binance_enabled: true` in the DB), applied the fix, confirmed the toggle now shows checked after reload. Screenshots attached below.

## Screenshots

Before (bug): Binance Pay switch unchecked after reload, despite `tenant_settings.binance_enabled = {"enabled": true}` in the DB.
After (fixed): Binance Pay switch correctly shows checked after reload.

(attached as a PR comment)

### DIFF
--- a/app/actions/admin/settings.ts
+++ b/app/actions/admin/settings.ts
@@ -35,7 +35,7 @@ export async function getSettings(category?: string): Promise<SettingsResponse> 
       const categoryPrefixes: Record<string, string[]> = {
         general: ['site_name', 'site_description', 'contact_email', 'support_email', 'timezone', 'maintenance_mode'],
         email: ['smtp_', 'email_'],
-        payment: ['stripe_', 'paypal_', 'lemonsqueezy_', 'solana_', 'currency', 'tax_rate', 'invoice_prefix', 'require_payment_approval', 'manual_payment_instructions'],
+        payment: ['stripe_', 'paypal_', 'lemonsqueezy_', 'solana_', 'binance_', 'currency', 'tax_rate', 'invoice_prefix', 'require_payment_approval', 'manual_payment_instructions'],
         enrollment: ['auto_enrollment', 'require_enrollment_approval', 'max_enrollments_per_user', 'allow_self_enrollment', 'enrollment_expiration_days', 'course_capacity_enabled'],
       }
       const keys = categoryPrefixes[category]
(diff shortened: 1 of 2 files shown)


guillermoscript:
## Visual verification

**Before** — Binance Pay toggle shows unchecked despite `tenant_settings.binance_enabled = {"enabled": true}` in the DB (Stripe/PayPal correctly show their saved state, Binance doesn't):

before

**After** — same DB row, same page, Binance Pay now correctly shows checked:

after

Issue #462, opened 2026-07-18
Auto-expire lapsed manual-transfer platform subscriptions (cron)


guillermoscript:
Part of #458 (Phase 1 — revenue leak).

## Problem
`app/api/cron/expire-subscriptions/route.ts` operates only on the student `subscriptions` table (lines 59, 96). Nothing ever expires `platform_subscriptions` rows with `payment_method = 'manual_transfer'` or downgrades `tenants.plan`. Stripe-paying schools auto-downgrade to free via `customer.subscription.deleted`, but a school that paid once by bank transfer **keeps its paid plan forever** after the period lapses, unless a super admin notices.

## Wanted
- Extend the cron (or add a dedicated one) to find lapsed manual platform subscriptions (`current period end + grace window < now`, no pending `platform_payment_requests` renewal) → set sub `expired`, `tenants.plan = 'free'`, `billing_status = 'free'`, reset `revenue_splits` (read the free plan row — don't hardcode 10/90 like `platform-webhook/route.ts:~222`).
- Email school admins before (renewal reminder — `requestManualRenewal` card already exists in `billing-dashboard-client.tsx:146-180`) and on downgrade.

## Acceptance
- [ ] Lapsed manual platform subscription auto-downgrades to free after the grace window.
- [ ] Pending renewal request pauses the downgrade.
- [ ] Reminder + downgrade notifications sent.

guillermoscript:
🤖 An AI agent (Claude Code, driven by @guillermoscript) is picking this issue up. Plan below — corrections welcome.

## Root cause / current behavior

The premise needs one correction: a downgrade path for lapsed manual platform subscriptions **does exist and runs** — `handle_manual_subscription_expiry()` (migration `20260221020000_subscription_expiry_cron.sql`), scheduled via **pg_cron** in the cloud DB (`cron.job` id 2, daily 02:00 UTC, verified succeeding through 2026-07-18). It marks lapsed `manual_transfer` subs `past_due` with a 7-day grace window, then downgrades (`status='canceled'`, `tenants.plan='free'`, `billing_status='free'`, revenue split reset).

What's actually broken, per the acceptance criteria:

1. **No pending-renewal pause** — the SQL function downgrades after grace even if the school already filed a `platform_payment_requests` renewal that's awaiting super-admin confirmation.
2. **No notifications** — pure SQL can't email; schools get zero warning before or after the downgrade.
[...]

PR #519, opened 2026-07-25
fix(ops): schedule /api/cron/* from GitHub Actions and reconcile cutoff on usage change (#513)

### LINKED ISSUE #513: Access-cutoff reconciliation depends on a cron that may never run on Dokploy (#494 follow-up)
Follow-up to #493 / #494 — found while verifying the epic's delivered state (2026-07-24).

## Problem

`reconcileAccessCutoff()` is the only thing that ever sets `tenants.access_cutoff_at`. For a tenant that simply grows past its plan limits (no plan-change event), the daily cron is the sole trigger — and that cron is declared only in `vercel.json`, while production runs on Dokploy.

## Evidence

**Call sites of `reconcileAccessCutoff`:**

- `lib/billing/downgrade-tenant.ts` (plan-change event)
- `lib/payments/platform-plan-change.ts` (plan-change event)
- `app/actions/admin/billing.ts` (plan-change event)
- `app/api/cron/enforce-plan-limits/route.ts` (the only organic-growth path)

Notably absent: `app/actions/join-school.ts` (new student joins) and `app/actions/teacher/courses.ts` (new course created) — the two places that already run soft plan-limit checks and where usage actually crosses a limit.

**Scheduling:** `/api/cron/enforce-plan-limits` is registered in `vercel.json` at `0 3 * * *`, alongside six other cron routes. `docs/DEPLOYMENT.md` is the Dokploy guide and documents `CRON_SECRET` but no scheduler; `docs/OPERATIONS_GUIDE.md:218` still describes "Configure Vercel Cron". No GitHub Actions workflow hits `/api/cron/*`. The pg_cron job for platform expiry was deliberately unscheduled in `supabase/migrations/20260719130000_platform_expiry_http_cron.sql` in favour of the HTTP route.

## Impact

If nothing is calling these routes on the Dokploy host, then: no cutoff is ever scheduled for organically over-limit tenants (#494 never fires for them), and platform subscriptions are never expired (#462's route never runs) — the whole non-payment lifecycle is inert in production. This is not specific to #494; it affects all seven cron routes.

## Wanted

1. Confirm whether a Dokploy scheduled task / host crontab is actually hitting `/api/cron/*` with `CRON_SECRET`. If not, add one and document it in `docs/DEPLOYMENT.md` next to the env vars.
2. Call `reconcileAccessCutoff()` from `join-school.ts` and `teacher/courses.ts` so a limit crossing is caught at the moment it happens, with the cron as a safety net rather than the only mechanism.
3. Fix `docs/OPERATIONS_GUIDE.md:218-223`, which still tells the operator to configure Vercel Cron.

guillermoscript:
Closes #513.

## What was actually wrong

I checked the premise against production before writing any code. Querying the Dokploy API for schedules on the LMS app (`schedule.list`, `wTNg0pha2N1VUYgzeB8-a`), on the MCP app, and at `dokploy-server` scope all return `[]`, and `server.all` is empty so there is no separate host holding a crontab either.

**There is no scheduler in production. All seven `/api/cron/*` routes have never run there.**

That is broader than the access-cutoff symptom the issue was filed for. `tenants.access_cutoff_at` is only written by `reconcileAccessCutoff()`, and five of its six call sites hang off a plan-change event; the sixth — `enforce-plan-limits` — is the only path that fires for a tenant that grows over its limits organically, and it's declared exclusively in `vercel.json`, which Dokploy does not read. The same silence applies to platform-subscription expiry (#462), the daily digest, league rollover, and both payment reconcilers.

Two corrections to the issue's evidence, both of which changed the shape of the fix:

1. **`docs/DEPLOYMENT.md` was not silent on scheduling.** It already had a `### 3.5 Cron Jobs` section with a host-crontab example — but listing only 4 of the 8 routes, and nobody ever ran it. So this was a "complete it and make it executable" job, not a "write it from scratch" job.
2. **There are 8 route directories under `app/api/cron/` but only 7 entries in `vercel.json`.** `solana-pull` has never been scheduled anywhere despite its own docstring saying it should run hourly. See "Deliberately out of scope" below.

## Changes

**1. `.github/workflows/cron.yml` (new)** — schedules all seven routes, one `on.schedule` entry per distinct cadence already declared in `vercel.json`, with a `case` block mapping `github.event.schedule` to the route(s) that tick belongs to. Auth reuses the `Authorization: Bearer $CRON_SECRET` header every route already implements, so no application code changed to support this. A `workflow_dispatch` route picker makes any single job runnable on demand — that's what makes this verifiable at review time rather than something you find out about in 24 hours.

I chose GitHub Actions over a Dokploy scheduled task deliberately: it lives in the repo, it's reviewable in this PR, and a reviewer can verify it without prod console access. Its weaknesses are real and are documented in the file's header comment and in DEPLOYMENT.md — GitHub delays scheduled runs under load (dropping high-frequency ones first, so the `*/10` reconcilers are the most affected), and it disables scheduled workflows after 60 days of repository inactivity. Neither is disqualifying for daily billing sweeps on an active repo, but the operator now has the Dokploy scheduled task documented as the belt-and-braces alternative.

`vercel.json` is left as-is — harmless, and removing it would strand the project if it's ever deployed to Vercel.

**2. `reconcileAccessCutoff()` at the two usage-crossing sites** — `app/actions/join-school.ts` after the `tenant_users` insert succeeds, and `app/actions/teacher/courses.ts` after the `courses` insert succeeds. Both already hold an admin client, which is what the function needs. Both calls are wrapped in `try/catch` that logs and swallows: reconciliation (or an email failure inside it) must never fail a join or a course creation that already succeeded. The function is idempotent by design — it only schedules when `violations.length > 0 && !currentCutoffAt` — so re-entry from the cron is safe.

**Being straight about what this second change buys, because the issue slightly overstates it: it is defense-in-depth, not a live bug fix.** `computePlanLimitViolations` compares with strict `>`, while both pre-flight checks block at `>=` (`join-school.ts` at `count >= maxStudents`, `checkCourseLimit` at `currentCount < limit`). A legitimate join or course creation therefore stops *at* the limit and cannot produce a violation. The value is in the drift between the duplicated checks — `join-school.ts` falls back to a hardcoded `50` when the `platform_plans` row is missing or inactive, and `checkCourseLimit` counts *all* courses while `countTenantUsage` counts only non-archived ones. Where those disagree a crossing can slip past the pre-flight, and this catches it at the moment it happens instead of up to 24h later. It also clears a stale cutoff the moment a tenant drops back under its limit.

**3. Docs** — `docs/OPERATIONS_GUIDE.md` §4d was telling the operator to "Configure Vercel Cron" and named only `expire-subscriptions`; it now describes the actual scheduler, its two required repo settings, and how to verify it. `docs/DEPLOYMENT.md` §3.5 now presents three mutually-exclusive options (Actions / Dokploy scheduled task / host crontab) with all seven routes listed in each, and states plainly that running two of them means every job fires twice.

## Required before this does anything

**This PR does not become effective on merge.** Two repository settings must be added under **Settings → Secrets and variables → Actions**:

| Kind | Name | Value |
|---|---|---|
| Secret | `CRON_SECRET` | Must match the `CRON_SECRET` env var already set on the Dokploy app |
| Variable | `CRON_BASE_URL` | `https://preciopana.com` |

The workflow fails loudly if either is missing rather than silently no-opping, which is precisely the failure mode being fixed here. A `401` in the run log means the two `CRON_SECRET` values don't match.

## Testing

- `npm run test:unit` — **287/287 passing across 27 files** (283 before this branch, plus 4 new). New `tests/unit/access-cutoff-call-sites.test.ts` covers both call sites: that `reconcileAccessCutoff` is called exactly once with the tenant id, and that a *rejecting* `reconcileAccessCutoff` still lets `joinCurrentSchool()` return `{ success: true }` and `createCourse()` resolve with the course.
- **Mutation-tested, so those four aren't vacuous.** I stripped the `reconcileAccessCutoff` call from both call sites and re-ran: all 4 fail (`Test Files 1 failed`, `Tests 4 failed`). Restored and re-verified the tree is clean.
- `npx tsc --noEmit` — no errors in any changed file.
- `npx eslint app/actions/join-school.ts app/actions/teacher/courses.ts` — clean.
- `npm run build` — passes.
- Workflow YAML parsed with a strict parser (6 unique schedules, no duplicates, all 7 `vercel.json` routes covered, 8 dispatch options). The `case` block was extracted and executed against all six schedule strings plus one unmapped string, confirming each maps to the right route and that an unmapped schedule exits non-zero instead of silently doing nothing. `actionlint` is not installed locally, so the workflow has not been through it — worth a glance from the reviewer.

### QA script — the one check that proves this works

Everything above verifies the code; only this verifies the *mechanism*, and it can only run after the two settings above are added and this is merged to `master` (GitHub only offers `workflow_dispatch` for workflows present on the default branch):

```bash
gh workflow run cron.yml -f route=enforce-plan-limits
gh run watch
```

Expect `HTTP 200` and a body like `{"success":true,"scheduled":0,"cleared":0,"none":N,"errors":0}` where `N` is your tenant count. `401` → the `CRON_SECRET` values don't match. Any other code → the route or the base URL is wrong.

Then confirm the schedules actually fire: check the Actions tab ~24h later for a green run on the `0 3 * * *` tick.

## Deliberately out of scope

Both worth filing as follow-ups rather than smuggling in here:

- **`solana-pull` is still not scheduled.** It submits on-chain USDC transfers for native Solana subscriptions and has never had a scheduler in `vercel.json` either. Putting a money-moving job on a timer deserves its own reviewed change, so it's available under `workflow_dispatch` only and documented as intentionally unscheduled.
- **Three separate plan-limit implementations.** `lib/billing/plan-limits.ts`, `join-school.ts`'s inline check, and `checkCourseLimit()` each compute limits independently with different fallbacks, different comparison operators, and different course-count semantics. Consolidating them onto `lib/billing/plan-limits.ts` is the real fix for the drift that change #2 above merely compensates for.

No UI surface changed, so there are no screenshots.

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -1,0 +1,137 @@
+# Scheduled invocation of the app's /api/cron/* routes (issue #513).
+#
+# Production runs on Dokploy, which does not read `vercel.json` — so the seven
+# cron routes declared there had no scheduler at all and every non-payment
+# lifecycle job (access-cutoff enforcement, platform-subscription expiry, daily
+# digest, league rollover, payment reconciliation) was inert in production.
+#
+# Each route authenticates with the same `Authorization: Bearer $CRON_SECRET`
+# header it already implements, so no application code changed to support this.
+#
+# Required repository settings (Settings → Secrets and variables → Actions):
+#   secret   CRON_SECRET    — must match the CRON_SECRET env var on the Dokploy app
+#   variable CRON_BASE_URL  — the production origin, e.g. https://preciopana.com
+# Both are checked at runtime; a missing one fails the run loudly rather than
+# silently doing nothing, which is the failure mode this workflow exists to fix.
+#
+# Caveats worth knowing before relying on this:
+#   * GitHub delays scheduled runs under load, and drops high-frequency ones
+#     first — the */10 reconcilers are the most affected. If exact cadence
+#     matters for those, prefer a Dokploy scheduled task (see docs/DEPLOYMENT.md).
+#   * GitHub disables scheduled workflows after 60 days with no repository
+#     activity. Re-enable from the Actions tab if that happens.
+#   * Run ONE mechanism. If a Dokploy schedule is added later, remove the
+#     corresponding schedule here so routes don't fire twice.
+
+name: Scheduled cron routes
+
+on:
+  schedule:
+    # Keep these in sync with vercel.json and with the case block below.
+    - cron: '*/10 * * * *' # solana-reconcile + binance-personal-reconcile
+    - cron: '0 * * * *' # daily-digest — hourly by design (each tenant sends at its own local hour)
+    - cron: '0 0 * * *' # expire-subscriptions
+    - cron: '0 1 * * 1' # league-rollover (pg_cron also runs it Mondays 00:05; this is the fallback)
+    - cron: '0 2 * * *' # expire-platform-subscriptions
+    - cron: '0 3 * * *' # enforce-plan-limits
+  workflow_dispatch:
+    inputs:
+      route:
+        description: 'Cron route to run now'
+        required: true
+        type: choice
+        options:
+          - enforce-plan-limits
+          - expire-platform-subscriptions
+          - expire-subscriptions
+          - daily-digest
+          - league-rollover
+          - solana-reconcile
+          - binance-personal-reconcile
+          # Never scheduled: solana-pull submits on-chain USDC transfers and has
+          # never had a scheduler in vercel.json either. Available here for
+          # manual runs only — putting it on a timer needs its own review.
+          - solana-pull
+
+permissions:
+  contents: read
+
+concurrency:
+  # Serialize per trigger so a slow run can't overlap the next tick.
+  group: cron-${{ github.event.schedule || inputs.route }}
+  cancel-in-progress: false
+
+jobs:
+  invoke:
+    name: Invoke cron route(s)
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - name: Resolve routes for this trigger
+        id: resolve
+        env:
+          SCHEDULE: ${{ github.event.schedule }}
+          DISPATCH_ROUTE: ${{ inputs.route }}
+        run: |
+          set -euo pipefail
+          if [ -n "${DISPATCH_ROUTE:-}" ]; then
+            routes="$DISPATCH_ROUTE"
+          else
+            case "${SCHEDULE:-}" in
+              '*/10 * * * *') routes='solana-reconcile binance-personal-reconcile' ;;
+              '0 * * * *')    routes='daily-digest' ;;
+              '0 0 * * *')    routes='expire-subscriptions' ;;
+              '0 1 * * 1')    routes='league-rollover' ;;
+              '0 2 * * *')    routes='expire-platform-subscriptions' ;;
+              '0 3 * * *')    routes='enforce-plan-limits' ;;
+              *)
+                echo "::error::Schedule '${SCHEDULE:-}' has no route mapping. Add it to the case block in .github/workflows/cron.yml."
+                exit 1
+                ;;
+            esac
+          fi
+          echo "Resolved routes: $routes"
+          echo "routes=$routes" >> "$GITHUB_OUTPUT"
+
+      - name: Invoke
+        env:
+          CRON_SECRET: ${{ secrets.CRON_SECRET }}
+          CRON_BASE_URL: ${{ vars.CRON_BASE_URL }}
+          ROUTES: ${{ steps.resolve.outputs.routes }}
+        run: |
+          set -euo pipefail
+
+          if [ -z "${CRON_BASE_URL:-}" ]; then
+            echo "::error::Repository variable CRON_BASE_URL is not set (expected e.g. https://preciopana.com)."
+            exit 1
+          fi
+          if [ -z "${CRON_SECRET:-}" ]; then
+            echo "::error::Repository secret CRON_SECRET is not set. It must match the CRON_SECRET env var on the Dokploy app."
+            exit 1
+          fi
+
+          base="${CRON_BASE_URL%/}"
+          failed=0
+
+          for route in $ROUTES; do
+            url="$base/api/cron/$route"
+            echo "--- GET $url"
+            # The secret is passed via env into the header and never printed;
+            # GitHub also masks it in logs. Body is capped so a large digest
+            # response can't flood the log.
+            code="$(curl -sS -o /tmp/cron-body.txt -w '%{http_code}' \
+              --max-time 600 --retry 2 --retry-delay 15 --retry-connrefused \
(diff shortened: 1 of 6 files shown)


guillermoscript:
Flagging from the #600 audit (provider-agnostic platform billing): this PR is the **blocking prerequisite** for the money-loop MVP acceptance in #605, not just for #513.

Without a scheduler, `expire-platform-subscriptions` and `enforce-plan-limits` never run, so the entire lapse → grace → downgrade → access-cutoff half of school billing is untestable on a real deploy — which is exactly what #605 has to sign off. Your finding that `schedule.list` returns `[]` at every Dokploy scope is cited in #600 as the reason.

Currently draft + CONFLICTING. Needs a rebase onto master, a merge, and the two repo settings (`CRON_SECRET` secret, `CRON_BASE_URL` variable). Please don't let this get reimplemented from scratch — #605 explicitly says merge this rather than rebuild it.

guillermoscript:
Rebased onto `master` (98 commits of drift) and resolved for current `HEAD`.

**Conflict resolution — `app/actions/teacher/courses.ts`**

Only one file conflicted, in its import block. Since this branch was cut, master rewrote `checkCourseLimit` to use `countTenantUsage` / `getTenantPlanLimits` from `lib/billing/plan-limits`, and added `reconcileAccessCutoffSafely` — a wrapper that is exactly "call `reconcileAccessCutoff`, catch, log". Three changes followed from that:

- Dropped the `PLAN_LIMITS_FALLBACK` constant this branch reintroduced. Master's rewritten `checkCourseLimit` does not use it, and `lib/billing/plan-limits.ts` documents its deletion as deliberate — a missing plan row already fails open there.
- Replaced both hand-rolled `try/catch` blocks (in `createCourse` and `joinCurrentSchool`) with `reconcileAccessCutoffSafely`. They were the only two places in the codebase still reimplementing that wrapper by hand; `archiveCourse` and `deleteCourse` in the same file already use it.
- Updated `tests/unit/access-cutoff-call-sites.test.ts` to match. The mock now supplies a faithful copy of the wrapper over a mocked `reconcileAccessCutoff`, so the two non-blocking tests still exercise a real rejection being swallowed rather than asserting that a no-op mock does nothing. The builder mock also gained `.neq()` and a course row count, which master's `countTenantUsage` now needs.

**Verification**

- `npm run test:unit` — 797 passed / 61 files
- `npm run typecheck` — clean
- `npx eslint` on the changed files — clean
- `.github/workflows/cron.yml` parses, and its route list still matches `app/api/cron/` exactly: 7 scheduled plus `solana-pull` on `workflow_dispatch` only

**Before this is switched on**

The workflow is being merged **disabled**, because the two repository settings it needs do not exist yet and a scheduled run without them fails loudly every 10 minutes:

1. Secret `CRON_SECRET` — must match the `CRON_SECRET` env var on the Dokploy app
2. Variable `CRON_BASE_URL` — the production origin

Once both are set, `gh workflow enable cron.yml` turns it on, and a `workflow_dispatch` run of `expire-platform-subscriptions` should return 200 rather than 401. Tracked on #605.

Issue #498, opened 2026-07-24
No refund/chargeback clawback for manual payouts already marked paid


guillermoscript:
Part of #493 (Phase 2 — payout money-accuracy).

## Problem
`markPayoutPaid` is a one-way ratchet — once a payout is recorded, nothing re-checks whether the underlying `transactions` it was based on later flip to `refunded`/`failed`. If a PayPal sale gets refunded after you've already paid the school their share, `/platform/payouts` has no way to reflect that you're now effectively owed money back — it'll just show $0 owed going forward for that tenant, silently absorbing the loss with no record of it.

## Wanted
When computing `netOwed` in `computeOwedBalances`, subtract refunded-transaction amounts (scaled by the split, same as any other transaction) from `grossCollected`/`grossOwed`, so a refund after payout is reflected as a negative adjustment on the *next* payout cycle rather than disappearing. Surface it in the UI (e.g. a "clawback" line) rather than silently netting it into `netOwed`.

## Acceptance
- [ ] A refund on a transaction that was already included in a paid-out payout reduces what's owed on the next cycle.
- [ ] The clawback is visible on the payouts page, not just silently absorbed into the running total.
- [ ] Unit test: transaction refunded after being paid out → next `computeOwedBalances` call reflects the clawback.

guillermoscript:
An AI agent is picking this up as part of epic #493's Phase 2 follow-through.

Fixed in PR #506: transactions with `status = 'refunded'` are now included in the payouts-owed calculation instead of being excluded from the query entirely. Their scaled amount (using the same split-percentage logic as any other transaction) is surfaced as a distinct `clawback` figure on `/platform/payouts`, both in a per-currency banner and a per-row column, and it reduces `netOwed` on the next cycle rather than silently vanishing.

Unit tests cover: refund-after-payout produces a visible clawback and reduces netOwed; clawback is reported even with nothing else outstanding; a refund in one currency doesn't clawback a balance in another; clawback uses the transaction's own snapshotted split (issue #496 semantics), not the tenant's current one.

guillermoscript:
Closed by #506, merged into `master`.

Refunded transactions no longer silently vanish from the payout calculation — they're now included, scaled by the split, and surfaced as a distinct clawback figure that reduces `netOwed` on the next cycle. Visible both in a per-currency banner and a per-row column on `/platform/payouts`. Full test record and live-verification screenshot are on the PR.

guillermoscript:
[...]

Issue #493, opened 2026-07-24
EPIC: Payout & non-payment enforcement integrity — revenue-split accuracy and edge cases


guillermoscript:
# EPIC: Payout & non-payment enforcement integrity — revenue-split accuracy and edge cases

## Summary

We just shipped manual payout tracking (`/platform/payouts` — tracks what the platform owes each school for PayPal/Binance Pay/Lemon Squeezy sales, since those providers settle 100% into the platform's own account with no automatic split) and live-verified it end-to-end (PayPal $10 sale → $8.00 owed at 80% school split → Mark as Paid → $0.00, `payouts` row inserted correctly).

Auditing what surrounds it — specifically "what happens if a school just stops paying us" — surfaced two separate classes of real gaps, code-verified 2026-07-24:

1. **Non-payment has no functional consequence.** A school that stops paying keeps 100% of its existing students/courses/access forever — there is no lockout, no grandfathering decision, just... nothing happens to existing usage. Only *new* signups/course creation get blocked.
2. **The payout math itself has money-accuracy bugs** — revenue split isn't time-sliced (plan changes retroactively reprice historical payouts), transaction amounts get summed across currencies with no grouping, and there's no clawback path if a paid-out transaction is later refunded.

None of this blocks what's already shipped — it's an accounting/enforcement hardening pass before this is trusted for real money at scale.

---

### 1.1 Downgrade-to-free never touches existing courses/students/entitlements 🔴
- `downgradeTenantToFree()` (`lib/billing/downgrade-tenant.ts:33-59`) — triggered by Stripe's `customer.subscription.deleted` webhook (`app/api/stripe/platform-webhook/route.ts:171`) or the `expire-platform-subscriptions` cron (`app/api/cron/expire-platform-subscriptions/route.ts:184,205`) — only ever writes `platform_subscriptions.status`, `tenants.plan/billing_status/billing_period_end`, and `revenue_splits`. It never queries or writes `courses`, `enrollments`, or `entitlements`.
- Course/lesson access is gated exclusively by `has_course_access()` (`supabase/migrations/20260516120000_entitlements_table.sql:64-78`), which checks only `entitlements.status='active'` + expiry — no reference to `tenants.plan` or any limit, ever.
[...]

PR #521, opened 2026-07-25
fix(payments): stop double-subtracting refunds from payout balances (#511)

### LINKED ISSUE #511: Payout clawback double-subtracts refunds and claws back sales never paid out (#498 follow-up)
Follow-up to #493 / #498 — found while verifying the epic's delivered state (2026-07-24).

## Problem

`computeOwedBalances` (`lib/payments/payouts-owed.ts:113-128`) subtracts a refund's school share twice, and subtracts it even when the underlying sale was never paid out. Both cases understate what the school is owed.

## A — double subtraction

A refunded transaction is excluded from `grossOwed` (the `continue` at line 121) and also subtracted again as `clawback`, since `netOwed = grossOwed - alreadyPaid - clawback`.

Worked example, 80% split: 1000 successful plus 100 refunded, where the refunded sale had already been paid out (80).

- computed: grossOwed 800, alreadyPaid 80, clawback 80, netOwed 640
- expected: 80% of net sales (800) minus what has already been paid (80) = 720

## B — clawback applied to refunds that were never paid out

The module has no notion of which transactions a given payout covered, so every refund is clawed back unconditionally.

Same figures, but with no payout ever recorded:

- computed: grossOwed 800, alreadyPaid 0, clawback 80, netOwed 720
- expected: 800, since nothing was ever paid out for the refunded sale

## Why this passed review

`tests/unit/payouts-owed.test.ts:185-202` encodes the current behavior as intended. Its fixture (100 successful, 100 refunded, 80 paid) happens to land on the right answer only because `Math.max(..., 0)` clamps -80 up to 0. Asymmetric amounts would have surfaced it.

## Wanted

- Pick one representation and keep it: either leave refunded sales in `grossOwed` and let `clawback` be the only subtraction, or exclude them from `grossOwed` and drop the clawback term. Not both.
- Only claw back refunds plausibly covered by a recorded payout. Comparing `payouts.paid_at` against `transactions.created_at` is enough for a first cut; a `covered_through` column on `payouts` would make it exact.
- Add unit tests with asymmetric amounts so the clamp cannot mask a sign error again.

guillermoscript:
Closes #511.

Follow-up to #493 / #498. A refunded sale was being subtracted from a school's payout balance twice, and refunds that had never been paid out were clawed back anyway. Both understated what the platform owes.

## What was wrong

`computeOwedBalances` carried two mutually exclusive representations of a refund at the same time. The transaction loop excluded refunded sales from `grossCollected`/`grossOwed`, and then the balance builder subtracted their scaled share *again* as `clawback`:

```ts
const netOwed = Math.max(grossOwed - alreadyPaid - clawback, 0)
```

The second subtraction was never needed in the first place. `alreadyPaid` is the **all-time** sum of recorded payouts, so it still contains whatever was paid out for a sale that later flipped to `refunded`. Excluding that sale from `grossOwed` while leaving its payout inside `alreadyPaid` recovers the overpayment on its own — there is nothing left to claw back.

Separately, the module had no notion of *when* anything happened, so it could not distinguish a refund whose payout had already gone out from one that was never paid. It clawed back both.

## What changed

**1. One representation.** `netOwed` is now `max(grossOwed - alreadyPaid, 0)`. Refunded sales leaving `grossOwed` is the only place a refund moves the balance.

**2. `clawback` becomes reporting-only, and is gated on plausibility.** It answers "how much of what we already paid this school was for sales that have since been refunded?" — the number an operator needs to understand why a balance dropped, not a second deduction. A refund contributes only when a payout in the **same currency** settled up to a point at or after the sale. Refunds that were never paid out contribute nothing.

**3. Dates wired through, no migration needed.** `PlatformSettledTxn` gains `transactionDate`, `ManualPayoutRecord` gains `coveredThrough`.

- The column is `transactions.transaction_date` — that table has **no `created_at`**, so the issue's suggested comparison needed adjusting.
- `coveredThrough` reads `payouts.period_end ?? paid_at ?? created_at`. `period_end` is already in the schema and is exactly the `covered_through` the issue wished for; it is simply null for every manually recorded payout today, which falls back to `paid_at`. So the "exact" version the issue floated needs no new column — only for `markPayoutPaid` to start stamping a period, which I left alone as out of scope.
- Dates are compared as parsed instants, not strings: `paid_at` and `period_end` can reach us as `…Z` or `…+00:00` depending on their source, and `'+00:00' < 'Z'` lexicographically would silently invert the comparison.
- A missing or unparseable date on either side answers "not covered". An unjustifiable clawback is worse than a missing one.

**4. UI copy on `/platform/payouts`.** The banner and the column used to read as a deduction the viewer should apply themselves, and the cell carried a minus sign. It is no longer a deduction, so the column is now "Of which refunded", the value has no sign, and the banner says the amount is already netted out and must not be collected again.

## Scope notes

- **The `Math.max(…, 0)` floor stays.** A school overpaid beyond its outstanding balance still reads `0`, not a negative — only a payout row moves money, and this view should not invent a reverse one. In that situation the unrecovered remainder is visible only through `clawback`, which is now the figure's most useful job. Documented in the module docstring.
- **The plausibility test is a heuristic**, and the docstring says so. A payout dated after a refunded sale is assumed to have covered it, even though the module has no record of which transactions a payout settled. That can overstate `clawback` when a payout deliberately skipped a sale. It cannot move `netOwed`, so the blast radius is a reported number, not money. Making it exact needs real payout-to-transaction linkage on `payouts` — happy to file that if you want it.

## Testing

- `npm run test:unit` — **313 passed, 27 files**. `tests/unit/payouts-owed.test.ts` is 26 tests, up from 20.
- `npm run typecheck` — clean.
- `npm run build` — clean.
- `npx eslint` on the 4 changed files — clean. The one warning (`'supabase' is assigned a value but never used`, `app/actions/platform/payouts.ts:12`) predates this branch and is untouched.

The four `#498` tests encoded the double-subtraction as intended behavior, so they are **rewritten rather than deleted** — the diff is the correction. Their old fixture (100 successful / 100 refunded / 80 paid) landed on the right answer only because the clamp lifted `-80` to `0`, so every new case uses asymmetric amounts where the result sits far from the floor:

| Case | Before | After |
|---|---|---|
| Refund already paid out (1000 + 100 refunded @ 80%, 80 paid) | `netOwed 640` | **`netOwed 720`**, `clawback 80` |
| Refund never paid out (same, no payout) | `netOwed 720`, `clawback 80` | **`netOwed 800`**, `clawback 0` |

New coverage: refund sold *after* the last payout, payout in a different currency, undated refund, undated payout, latest-payout-wins across several payouts, mixed ISO offset formats, and a fully-refunded/fully-paid school where the floor holds and `clawback` is the only trace.

## Browser QA

Live against local Supabase on `default.lvh.me:3005` as super admin (`owner@e2etest.com`), seeded to the issue's worked example exactly — Default School at 80%, PayPal 1000 successful + 100 refunded, one manual payout of 80 recorded after the sale. Fixture rows deleted afterwards.

**Before — `master`, the bug reproducing at $640.00:**

Payouts page on master showing Currently Owed 640.00

**After — the correct $720.00, with the reworded clawback banner and column:**

Payouts page with the fix showing Currently Owed 720.00

**After, with the payout removed — case B: $800.00, no clawback banner, `—` in the refunded column:**

Payouts page with no payout recorded showing Currently Owed 800.00

No GIF: there is no interactive flow here — the change is one number on a static server-rendered table, and the three states above are the evidence.

### DIFF
--- a/app/[locale]/platform/payouts/page.tsx
+++ b/app/[locale]/platform/payouts/page.tsx
@@ -140,8 +140,9 @@ export default async function PlatformPayoutsPage() {
           data-testid="payouts-clawback-banner"
         >
           <span className="font-medium">Refund clawback: {formatByCurrency(totalClawbackByCurrency)}.</span>{' '}
-          Some already-paid-out transactions were later refunded — this amount reduces what&apos;s
-          currently owed rather than being paid again.
+          That much of what you already paid out was for sales that have since been refunded. It is
+          already netted out of the balances below, so don&apos;t collect it again — nothing extra to
+          do here.
         </div>
       )}
 
@@ -163,7 +164,7 @@ export default async function PlatformPayoutsPage() {
                     <th className="pb-2 text-right font-medium">Collected</th>
                     <th className="pb-2 text-right font-medium">School %</th>
                     <th className="pb-2 text-right font-medium">Paid so far</th>
-                    <th className="pb-2 text-right font-medium">Clawback</th>
+                    <th className="pb-2 text-right font-medium">Of which refunded</th>
                     <th className="pb-2 text-right font-medium">Owed</th>
                     <th className="pb-2 text-right font-medium"></th>
                   </tr>
@@ -187,8 +188,14 @@ export default async function PlatformPayoutsPage() {
                       </td>
                       <td className="py-2.5 text-right tabular-nums">
                         {r.clawback > 0 ? (
-                          <span className="font-medium text-red-600 dark:text-red-400">
-                            −{money(r.clawback, r.currency)}
+                          // Not a deduction column: this is the slice of "Paid so far"
+                          // that covered sales later refunded. "Owed" already accounts
+                          // for it (#511), so it carries no minus sign.
+                          <span
+                            className="font-medium text-red-600 dark:text-red-400"
+                            title="Part of what was already paid out, for sales later refunded. Already reflected in Owed."
+                          >
+                            {money(r.clawback, r.currency)}
                           </span>
                         ) : (
                           <span className="text-muted-foreground">—</span>
--- a/app/actions/platform/payouts.ts
+++ b/app/actions/platform/payouts.ts
@@ -29,10 +29,14 @@ export async function getPayoutsOwed(): Promise<TenantOwed[]> {
     admin.from('revenue_splits').select('tenant_id, school_percentage'),
     admin
       .from('transactions')
-      .select('tenant_id, payment_provider, amount, currency, school_percentage_snapshot, status')
+      .select('tenant_id, payment_provider, amount, currency, school_percentage_snapshot, status, transaction_date')
       .in('status', ['successful', 'refunded'])
       .in('payment_provider', PLATFORM_SETTLED_PROVIDERS),
-    admin.from('payouts').select('tenant_id, amount, currency').eq('payout_method', 'manual').eq('status', 'paid'),
+    admin
+      .from('payouts')
+      .select('tenant_id, amount, currency, period_end, paid_at, created_at')
+      .eq('payout_method', 'manual')
+      .eq('status', 'paid'),
   ])
 
   const schoolPercentageByTenant = new Map(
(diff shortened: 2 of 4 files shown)


guillermoscript:
Merged as `155a8834`.

Shipped: refunded sales now leave `grossOwed` and nothing else — `netOwed = max(grossOwed - alreadyPaid, 0)`. The `clawback` term is gone from the arithmetic and survives as a reporting-only figure, gated on a payout in the same currency having settled at or after the sale, so refunds that were never paid out no longer reduce anything. Verified live on `/platform/payouts` against the issue's own worked example: 640 → 720 with a payout recorded, 800 with none.

No migration was needed — `payouts.period_end` already served as the `covered_through` column the issue asked for, falling back to `paid_at` for the manually recorded payouts that leave it null.

Two things deliberately left out of scope, either of which is worth its own issue if you want them:

- **Exact clawback attribution.** The plausibility test is still a heuristic — a payout dated after a refunded sale is assumed to have covered it. Making it exact needs real payout-to-transaction linkage on `payouts`. It cannot move `netOwed`, so the exposure is a reported number only.
- **`markPayoutPaid` does not stamp a period.** It leaves `period_start`/`period_end` null, so every manual payout falls back to `paid_at`. Stamping a period at record time would make the comparison exact for all future payouts with no schema change.

Issue #496, opened 2026-07-24
Revenue split isn't time-sliced — plan changes retroactively reprice historical payouts


guillermoscript:
Part of #493 (Phase 2 — payout money-accuracy).

## Problem
`computeOwedBalances` (`lib/payments/payouts-owed.ts:48-83`) takes one **current** `schoolPercentage` per tenant and applies it to the **entire all-time sum** of that tenant's platform-settled transactions — it has no concept of "what split was in effect on the day of each sale."

Every plan change (upgrade, downgrade, the free-plan auto-downgrade, or a super-admin `forceTenantPlanChange`) rewrites `revenue_splits` for that tenant (`lib/billing/downgrade-tenant.ts:52-59`, `app/api/stripe/platform-webhook/route.ts:87-103`) — and the next payout calculation silently reprices *every historical transaction* at the new percentage, not just new ones.

Concrete failure: a school does $10k on the business plan (0% platform fee) then gets auto-downgraded to free (10% fee per `supabase/migrations/20260221020000_subscription_expiry_cron.sql:71-72`) — the payouts page then shows ~$1k less owed on that already-settled $10k than was actually agreed at the time of sale.

## Wanted
Snapshot the split percentage onto each `transactions` row at creation time (simplest), or maintain a `revenue_splits` history table keyed by effective date range. Compute owed amounts per-transaction using the split that was live when that transaction happened, never today's split applied retroactively.

## Acceptance
- [ ] A transaction's contribution to "owed" is fixed at the split rate in effect when it happened.
- [ ] A tenant plan change never retroactively changes already-computed owed amounts for past transactions.
- [ ] Unit test: tenant makes a sale at split A, plan changes to split B, verify owed reflects A for that sale and B only for sales after the change.

guillermoscript:
Verification pass on #493 (2026-07-24): the snapshot is written at exactly one call site (`app/api/payments/checkout/route.ts:234`) with no default or trigger. Safe today only because the other four transaction-insert paths leave `payment_provider` NULL. Hardening tracked as #512.

PR #525, opened 2026-07-25
fix(db): back the revenue-split snapshot with a trigger, not one call site (#512)

### LINKED ISSUE #512: Split snapshot is written at a single call site with no DB backstop (#496 follow-up)
Follow-up to #493 / #496 — found while verifying the epic's delivered state (2026-07-24).

## Problem

#496 fixed retroactive repricing by snapshotting `revenue_splits.school_percentage` onto each transaction. That snapshot is written in exactly one place, with no database-level backstop, so any future transaction-insert path silently reverts to the old (buggy) fallback behavior.

## Evidence

Repo-wide, `school_percentage_snapshot` is written at exactly one site:

- `app/api/payments/checkout/route.ts:234`

The other transaction-insert sites do not set it:

- `app/[locale]/(public)/checkout/actions.ts:54` (mock/dev product checkout)
- `app/[locale]/(public)/checkout/actions.ts:102` (mock/dev plan checkout)
- `app/actions/payment-requests.ts:359` (admin confirms manual/offline payment)
- `app/api/stripe/create-payment-intent/route.ts:141` (Stripe Connect)

Column state confirmed against the live schema: `school_percentage_snapshot` is nullable with no default, and `payment_provider` is likewise nullable with no default.

## Why it is not currently wrong

`getPayoutsOwed()` (`app/actions/platform/payouts.ts:33-34`) filters on `payment_provider IN (paypal, lemonsqueezy, binance)`. The four unsnapshotted paths leave `payment_provider` NULL (they write `payment_method` instead) or use Stripe Connect, so none of them reach the payout computation today. The correctness of #496 rests entirely on that coincidence.

## Risk

Any new or edited code path that inserts a transaction with a platform-settled `payment_provider` and forgets the snapshot falls back to `schoolPercentageByTenant` in `computeOwedBalances` — i.e. the tenant's *current* split applied retroactively, which is exactly the bug #496 closed. Nothing fails loudly; the number is just quietly wrong.

## Wanted

- Move the snapshot to a `BEFORE INSERT` trigger on `transactions` so it is set regardless of caller, keeping the app-layer write as a harmless override.
- Or, at minimum, add a `CHECK` / test that asserts every row with a platform-settled `payment_provider` carries a non-null snapshot, so the omission fails loudly instead of silently repricing.

DPS0340:
Closes #512.

## The gap is one step wider than the issue describes

The issue proposes a `BEFORE INSERT` trigger, on the premise that the four unsnapshotted insert sites are harmless today because they leave `payment_provider` NULL (they write `payment_method`) or use Stripe Connect, and `getPayoutsOwed()` filters on platform-settled `payment_provider` values.

That premise holds for the insert. It does not hold for the row's lifetime — `payment_provider` is also written by **UPDATE**, after the row exists:

- `lib/payments/webhook-dispatch.ts:205` — `.update({ status: 'successful', payment_provider: provider })`
- `app/api/stripe/webhook/route.ts:406` — same shape for Stripe subscriptions

`dispatchBillingEvent` is the shared activation path for PayPal (`app/api/payments/paypal/capture/route.ts`), the unified provider webhook (`app/api/payments/webhook/[provider]/route.ts`), and Stripe. PayPal and Lemon Squeezy are exactly the providers carrying `settlesToPlatformAccount: true` — the ones the payout computation reads.

So a row can be inserted with `payment_provider` NULL, sail past a `BEFORE INSERT` trigger with nothing to do, and *then* be turned into a platform-settled transaction by a webhook — landing in the payout math with a NULL snapshot and falling back to the tenant's current split. The #496 bug, reached by a path the proposed trigger never sees.

Hence `BEFORE INSERT OR UPDATE`.

## What it does, and deliberately does not

The trigger fires only while `school_percentage_snapshot IS NULL` (a `WHEN` clause on the trigger itself, so it costs nothing on the overwhelming majority of writes) and fills it from `revenue_splits`, falling back to 80 when the tenant has no split row — matching `DEFAULT_SCHOOL_PERCENTAGE`, `create-payment-intent`'s existing default, and the seeded row.

- **An explicit app-layer value always wins.** The checkout route's write stays authoritative; this only fills gaps.
- **History is never re-stamped.** On UPDATE it acts only while the value is still NULL. Re-stamping a historical row with a newer split *is* the #496 bug, not a fix for it.
- **Existing NULLs are not backfilled.** Rows predating `20260724140000` are documented — in that migration and in `payouts-owed.ts` — to fall back to the tenant's current split. Backfilling would stamp today's split onto history and manufacture exactly the retroactive repricing #496 removed, while looking authoritative doing it. A documented fallback is the better failure mode.

**Not a CHECK constraint** (the issue's fallback suggestion): a CHECK cannot read `revenue_splits`, and one on `payment_provider IS NULL OR snapshot IS NOT NULL` would hard-fail the webhook UPDATE above for any legacy NULL-snapshot row a provider later activates — converting a quiet mispricing into a failed activation and a lost sale. Filling the value beats rejecting the row.

The existing `after_transaction_insert` / `after_transaction_update` triggers are both AFTER, so a BEFORE trigger composes with them without ordering concerns.

## Verification — please read before merging

**The trigger's behaviour is SQL and I could not execute it here.** This repo's vitest suite is `environment: 'node'` with no database, and I have no Supabase CLI or `.env.local` in this environment. I would rather say that plainly than imply coverage I do not have.

What I have provided instead:

- `supabase/snippets/verify_transaction_split_snapshot_backstop.sql` — executable verification covering all five cases, each printing PASS/FAIL, wrapped in a transaction that rolls itself back so it is safe against a populated database. Case 4 is the important one: it disables the trigger to create a genuinely NULL-snapshot row, then runs the webhook-shaped UPDATE and asserts the snapshot gets filled — the gap a `BEFORE INSERT` trigger would leave open. Case 5 changes the tenant's split and asserts an existing snapshot is *not* re-stamped.
- `tests/unit/transaction-split-snapshot-backstop.test.ts` — guards what can be checked statically: that the fallback percentage in the plpgsql cannot drift from `DEFAULT_SCHOOL_PERCENTAGE`, that the trigger covers UPDATE, that the `WHEN` guard is present, and that no backfill was introduced.
- `supabase/migrations/rollback/…down.sql`, per the convention in `20260724150000_content_entitlement_rls.down.sql`.

Suite-level: `npm run test:unit` 322 passed (28 files), `npm run typecheck` clean, `npx eslint` clean on changed TS. Repo CI green on my fork: https://github.com/DPS0340/lms-front/actions/runs/30148686141

A review pass against a real database would be worth doing before this merges.

### DIFF
--- a/app/api/payments/checkout/route.ts
+++ b/app/api/payments/checkout/route.ts
@@ -207,9 +207,18 @@ export async function POST(req: NextRequest) {
 
     // Snapshot the tenant's CURRENT revenue split onto this transaction (#496)
     // so a later plan change doesn't retroactively reprice it in the payouts
-    // computation — revenue_splits is super-admin-only under RLS, so this
-    // needs the admin client even though the rest of this route uses the
-    // user-scoped one.
+    // computation. The admin client is belt-and-braces, not a requirement: this
+    // comment used to say revenue_splits is "super-admin-only under RLS", which
+    // is wrong — the SELECT policy is `tenant_id = get_tenant_id() OR
+    // is_super_admin()`, so the user-scoped client could read it too (#512).
+    //
+    // Since #512 the DATABASE owns this column: the BEFORE INSERT trigger in
+    // 20260725110000_transaction_split_snapshot_backstop.sql recomputes it from
+    // the same table and ignores whatever we send, because the column is
+    // writable by any authenticated client (transactions RLS restricts rows, not
+    // columns) and a student-supplied value must not survive. This write is kept
+    // deliberately: it computes the identical number, and it keeps the rollback
+    // migration a safe lever — drop the trigger and this path still snapshots.
     const adminClientForSplit = createAdminClient()
     const { data: revenueSplit } = await adminClientForSplit
       .from('revenue_splits')
--- a/lib/payments/payouts-owed.ts
+++ b/lib/payments/payouts-owed.ts
@@ -14,10 +14,21 @@
  * snapshotted at insert time in `app/api/payments/checkout/route.ts` — see
  * issue #496), not the tenant's current split applied retroactively to the
  * whole sum. A plan change (which rewrites `revenue_splits`) therefore only
- * affects transactions created after the change. Transactions predating the
- * snapshot column (`schoolPercentageSnapshot: null`) fall back to the
- * tenant's current `schoolPercentage` — the same behavior this module had
- * before #496, so old data isn't retroactively wrong either way.
+ * affects transactions created after the change.
+ *
+ * That snapshot is also enforced in the database (issue #512): a
+ * BEFORE INSERT OR UPDATE trigger on `transactions` fills it from
+ * `revenue_splits` whenever a caller omits it, so a transaction-insert path
+ * that forgets — or a webhook that sets `payment_provider` on an existing row
+ * post-insert — cannot silently fall back to the current-split behaviour.
+ * The trigger never overwrites a non-NULL snapshot, so the app-layer write
+ * stays authoritative and history is never re-stamped.
+ *
+ * Transactions predating the snapshot column (`schoolPercentageSnapshot:
+ * null`) still fall back to the tenant's current `schoolPercentage` — the
+ * same behavior this module had before #496, so old data isn't retroactively
+ * wrong either way. They were deliberately not backfilled: stamping today's
+ * split onto history would manufacture the very repricing #496 removed.
  *
  * Balances are grouped PER CURRENCY (issue #497) — a tenant with both USD and
  * EUR sales owes two separate numbers, never one meaningless summed total.
--- a/lib/payments/payouts-owed.ts
+++ b/lib/payments/payouts-owed.ts
@@ -16,19 +16,31 @@
  * whole sum. A plan change (which rewrites `revenue_splits`) therefore only
  * affects transactions created after the change.
  *
- * That snapshot is also enforced in the database (issue #512): a
- * BEFORE INSERT OR UPDATE trigger on `transactions` fills it from
- * `revenue_splits` whenever a caller omits it, so a transaction-insert path
- * that forgets — or a webhook that sets `payment_provider` on an existing row
- * post-insert — cannot silently fall back to the current-split behaviour.
- * The trigger never overwrites a non-NULL snapshot, so the app-layer write
- * stays authoritative and history is never re-stamped.
+ * Since issue #512 that snapshot is OWNED BY THE DATABASE, not by the caller.
+ * `20260725110000_transaction_split_snapshot_backstop.sql` computes it from
+ * `revenue_splits` on INSERT and freezes it on UPDATE, so:
+ *
+ *   - a transaction-insert path that forgets it (four of the five do) can't
+ *     silently fall back to the current-split behaviour;
+ *   - a webhook that sets `payment_provider` on an existing row post-insert
+ *     (`lib/payments/webhook-dispatch.ts`) can't either;
+ *   - and no client can supply the number. `transactions` grants ALL to
+ *     `authenticated` and its RLS policies restrict which ROWS a user may
+ *     write, never which columns — so without the trigger a student could set
+ *     `school_percentage_snapshot: 100` on their own transaction and inflate
+ *     `grossOwed` below.
+ *
+ * The value never changes once written, so history is never re-stamped —
+ * re-stamping would be the #496 bug, not a fix for it.
  *
  * Transactions predating the snapshot column (`schoolPercentageSnapshot:
  * null`) still fall back to the tenant's current `schoolPercentage` — the
  * same behavior this module had before #496, so old data isn't retroactively
  * wrong either way. They were deliberately not backfilled: stamping today's
- * split onto history would manufacture the very repricing #496 removed.
+ * split onto history would manufacture the very repricing #496 removed. Such
+ * a row is snapshotted in exactly one situation — a provider activating it
+ * (its `payment_provider` changing), the one moment at which today's split is
+ * the honest answer for it.
  *
  * Balances are grouped PER CURRENCY (issue #497) — a tenant with both USD and
  * EUR sales owes two separate numbers, never one meaningless summed total.
--- a/supabase/migrations/20260725110000_transaction_split_snapshot_backstop.sql
+++ b/supabase/migrations/20260725110000_transaction_split_snapshot_backstop.sql
@@ -1,0 +1,90 @@
+-- Issue #512 — database backstop for the revenue-split snapshot.
+--
+-- #496 fixed retroactive repricing by snapshotting revenue_splits.school_percentage
+-- onto each transaction, but that snapshot is written at exactly ONE call site
+-- (app/api/payments/checkout/route.ts). Every other transaction-insert path
+-- omits it, and nothing fails when they do: computeOwedBalances() silently falls
+-- back to the tenant's CURRENT split, which is the #496 bug returning by a side
+-- door. The number is just quietly wrong.
+--
+-- WHY INSERT *OR UPDATE*, not INSERT alone.
+--
+-- The obvious backstop is BEFORE INSERT, on the reasoning that the unsnapshotted
+-- insert sites are harmless because they leave payment_provider NULL (they write
+-- payment_method instead) or use Stripe Connect, and getPayoutsOwed() filters on
+-- platform-settled payment_provider values. That reasoning holds for the insert
+-- and fails for the row's lifetime: payment_provider is also written by UPDATE,
+-- after the row exists —
+--
+--   lib/payments/webhook-dispatch.ts:205  .update({ status, payment_provider })
+--   app/api/stripe/webhook/route.ts:406   same shape, Stripe subscriptions
+--
+-- dispatchBillingEvent is the shared activation path for PayPal, the unified
+-- provider webhook (Lemon Squeezy et al), and Stripe. PayPal and Lemon Squeezy
+-- are precisely the settlesToPlatformAccount providers the payout computation
+-- reads. So a row can be inserted with payment_provider NULL, sail past a
+-- BEFORE INSERT trigger with nothing to do, and only then be turned into a
+-- platform-settled transaction by a webhook — arriving in the payout math with
+-- a NULL snapshot. Covering UPDATE closes that path.
+--
+-- WHAT IT DELIBERATELY DOES NOT DO.
+--
+-- 1. It never overwrites an existing snapshot. On UPDATE it acts only while the
+--    value is still NULL, so a historical transaction is never re-stamped with a
+--    newer split — re-stamping would BE the #496 bug, not a fix for it. The
+--    app-layer write therefore stays authoritative; this only fills gaps.
+--
+-- 2. It does not backfill existing NULLs. Rows predating 20260724140000 are
+--    documented (in that migration and in lib/payments/payouts-owed.ts) to fall
+--    back to the tenant's current split. Backfilling would stamp TODAY's split
+--    onto history and manufacture exactly the retroactive repricing #496
+--    removed — and it would look authoritative while doing it, which is worse
+--    than a documented fallback.
+--
+-- A CHECK constraint was the alternative offered in the issue. It cannot read
+-- revenue_splits, and a check like `payment_provider IS NULL OR snapshot IS NOT
+-- NULL` would hard-fail the webhook UPDATE above for any legacy NULL-snapshot
+-- row a provider later activates — converting a quiet mispricing into a failed
+-- activation and a lost sale. Filling the value beats rejecting the row.
+
+-- Mirrors DEFAULT_SCHOOL_PERCENTAGE in lib/payments/payouts-owed.ts, the 20%
+-- platform default in app/api/stripe/create-payment-intent/route.ts, and the
+-- row seeded by 20260216212440_create_revenue_infrastructure.sql. Used only
+-- when a tenant has no revenue_splits row at all.
+CREATE OR REPLACE FUNCTION public.set_transaction_split_snapshot()
+RETURNS trigger
+LANGUAGE plpgsql
+SECURITY DEFINER
+SET search_path TO 'public'
+AS $function$
+DECLARE
+  v_school_percentage numeric;
+BEGIN
+  -- Explicit app-layer value wins, always.
+  IF NEW.school_percentage_snapshot IS NOT NULL THEN
+    RETURN NEW;
+  END IF;
+
+  IF NEW.tenant_id IS NULL THEN
+    RETURN NEW;
+  END IF;
+
+  SELECT school_percentage INTO v_school_percentage
+  FROM revenue_splits
+  WHERE tenant_id = NEW.tenant_id;
+
+  NEW.school_percentage_snapshot := COALESCE(v_school_percentage, 80);
+
+  RETURN NEW;
+END;
+$function$;
+
+COMMENT ON FUNCTION public.set_transaction_split_snapshot() IS
+  'Backstop for #496 (issue #512): fills transactions.school_percentage_snapshot from revenue_splits when a caller omits it. Never overwrites a non-NULL snapshot, so historical rows are not repriced. Fires on UPDATE too, because payment_provider is set post-insert by the webhook activation path (lib/payments/webhook-dispatch.ts).';
+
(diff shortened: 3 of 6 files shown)


guillermoscript (dismissed):
Thanks for this — the analysis in the description is genuinely good, and the central insight is correct: `payment_provider` is written by UPDATE (`lib/payments/webhook-dispatch.ts:205`, `app/api/stripe/webhook/route.ts:406`), so a `BEFORE INSERT`-only trigger would leave that path open. I verified that against the code.

A few things I checked and confirm as correct, so they don't get re-litigated:

- `revenue_splits` has `CONSTRAINT revenue_splits_tenant_unique UNIQUE (tenant_id)` (`20260216212440`), so the un-`LIMIT`ed `SELECT … INTO` can't pick an arbitrary row.
- `SECURITY DEFINER` is genuinely needed, not cargo-culted: `revenue_splits` is admin/super-admin-only under RLS, and the checkout insert at `app/api/payments/checkout/route.ts:223` runs on the **user-scoped** client. Without DEFINER the `SELECT` would return nothing for a student and silently stamp the 80 fallback. `search_path` is pinned and there's no dynamic SQL.
- The `80` fallback matches `DEFAULT_SCHOOL_PERCENTAGE`, the `revenue_splits.school_percentage` column default, and the seeded row.
- The `WHEN` clause keeps this off the hot path, and the pre-existing `after_transaction_*` triggers are AFTER, so ordering is fine.
- `trigger_manage_transactions` no-ops on rows with NULL `product_id`/`plan_id`, so the verification snippet's claim about that is accurate.

Below, **[blocking]** = I'd like it addressed before merge.

---

## 1. [blocking] "An explicit app-layer value always wins" makes the backstop bypassable by the untrusted path

`transactions` grants `ALL` to `authenticated` (`20260126190500_lms_complete.sql:3692`), and RLS lets a user write their own rows with no column restriction (`20260313025318_rls_transactions.sql`):

```sql
```

So `school_percentage_snapshot` is directly client-writable. A student can `POST /rest/v1/transactions` with `school_percentage_snapshot: 100` and this trigger will preserve it by design — the value flows straight into `computeOwedBalances`' `grossOwed` and inflates what the platform owes that school. The rule that makes the trigger safe against re-stamping is the same rule that makes it defer to the one writer we don't trust.

The fix is cheap, because the trigger and the app read **the same source of truth**: the checkout route's value is just `revenue_splits.school_percentage` fetched over the wire a few lines earlier. So drop the "explicit wins" rule *on INSERT* and let the DB compute it unconditionally:

```sql
IF TG_OP = 'UPDATE' AND NEW.school_percentage_snapshot IS NOT NULL THEN
  RETURN NEW;   -- never re-stamp an existing row
END IF;
-- on INSERT, always compute from revenue_splits; ignore whatever the client sent
```

That keeps every property you argued for (no re-stamping, no backfill, app behaviour unchanged for the legitimate path) and removes the client's ability to inject the number at all. `app/api/payments/checkout/route.ts` can then keep or drop its own write — either way the values agree.

If you'd rather keep app-supplied values authoritative, the alternative is column-level privileges — but note a plain `REVOKE INSERT, UPDATE (school_percentage_snapshot) …` is a no-op while the table-level grant is held; you'd have to revoke table-level `INSERT, UPDATE` from `authenticated` and re-grant the explicit column list, which is a much larger blast radius. I'd take the trigger change.

*(Adjacent and pre-existing, not yours to fix here: the same policy lets a user insert `status: 'successful', payment_provider: 'paypal'` on their own row, which both self-enrolls them via `trigger_manage_transactions` and lands in the payout math. That deserves its own issue — mentioning it because it's the reason I'd not want the snapshot column to stay client-writable.)*

## 2. [blocking] The UPDATE branch is a lazy, non-deterministic partial backfill — which is the thing the description argues against

> **History is never re-stamped.** … **Existing NULLs are not backfilled.**

Both are true only in the narrow "never overwrite a non-NULL value" sense. With the trigger installed, every *new* row gets its snapshot at INSERT, so the UPDATE branch can only ever fire on **legacy NULL rows**. What it does to them is stamp **today's** split — permanently — at whatever arbitrary moment something first touches the row.

Concretely: a May transaction that gets flipped to `'refunded'` next week is frozen at next week's split, not the split in effect when it sold. That is the same retroactive-repricing error the description rejects backfilling for, just applied to an unpredictable subset of history and with no audit trail of which rows were converted.

The only UPDATE path the description actually justifies is provider activation. I'd narrow it to that. It needs splitting into two triggers, because `OLD` can't be referenced in a `WHEN` clause on a combined `INSERT OR UPDATE` trigger:

```sql
CREATE TRIGGER before_transaction_split_snapshot_insert
  BEFORE INSERT ON transactions
  FOR EACH ROW WHEN (NEW.school_percentage_snapshot IS NULL)
  EXECUTE FUNCTION public.set_transaction_split_snapshot();

CREATE TRIGGER before_transaction_split_snapshot_update
  BEFORE UPDATE ON transactions
  FOR EACH ROW
  WHEN (NEW.school_percentage_snapshot IS NULL
        AND NEW.payment_provider IS DISTINCT FROM OLD.payment_provider)
  EXECUTE FUNCTION public.set_transaction_split_snapshot();
```

If you'd rather keep the broad UPDATE, that's defensible — but then the description and the migration comment should say plainly that legacy NULL rows *are* backfilled with the current split, lazily and on touch, rather than stating the opposite.

## 3. [blocking] The verification snippet is not safe to run against a populated database

> wrapped in a transaction that rolls itself back so it is safe against a populated database

`ALTER TABLE transactions DISABLE TRIGGER …` takes a lock that blocks concurrent INSERT/UPDATE on the live payments table for the duration of the script, and if any statement errors the session can be left holding it. It also requires table ownership, so it won't run as `service_role` at all. The rollback protects *data*, not *availability*.

Please relabel it local-only (`npm run db:reset`, then run the file) and drop the "safe against a populated environment" claim. Everything else about the snippet is well constructed — case 4 in particular is the right test to have written.

## 4. [non-blocking] The unit tests are `grep`-over-SQL, and shouldn't read as coverage

All four assertions are regexes against the migration's *text*. They pass identically whether the trigger works, is broken, or was never applied to any database — and one of them (`not.toMatch(/UPDATE\s+transactions\s+SET/i)`) asserts that you didn't write something, which no future regression will trip. They're also brittle to harmless reformatting (`BEFORE INSERT OR\n  UPDATE` fails the match).

I'd keep the `DEFAULT_SCHOOL_PERCENTAGE`-drift one — that's a real duplication guard and the only one earning its place — and drop or clearly re-label the rest so CI green doesn't imply the trigger was exercised.

## 5. [non-blocking] Smaller things

- `supabase/snippets/` currently holds Supabase Studio dumps (`Untitled query 374.sql`). A hand-written verification script reads oddly there; `supabase/tests/` or alongside the migration would be clearer.
- The trigger now also stamps Stripe Connect and manual/offline transactions, where the money never touches the platform account and the snapshot is meaningless. Harmless today because `getPayoutsOwed()` filters on `PLATFORM_SETTLED_PROVIDERS`, but worth a line in the migration comment so a future reader doesn't take those values as significant.
- Minor info-disclosure consequence of `SECURITY DEFINER`: the function bypasses RLS on `revenue_splits` and writes the result onto a row the caller can `SELECT` back. Students can read their own transactions, so after this change a student can learn their school's revenue-split percentage — which the `"Admins can view own revenue split"` policy currently limits to admins. Low severity, but it's a new read path created by this PR.
- No `docs/MONETIZATION.md` / `docs/DATABASE_SCHEMA.md` note. Recent DB work in this repo has been landing with cloud deploy + verification evidence (e.g. `cd91aa33`, `c0f0b7f3`); worth matching.

## 6. Verification status

Appreciated that you flagged the gap plainly rather than implying coverage you didn't have — that's the right call. Recording where it stands so it isn't lost: **no DB verification has run yet on our side either.** No CI checks are reported on this branch (fork PR — workflows need maintainer approval), and I couldn't execute the snippet locally (Docker not running here). Before merge this needs `supabase db reset` + the snippet, and the five PASS lines pasted into the PR.

guillermoscript:
Followed up on this against a local database, and opened #527 with your commit as its base — the core insight (that `payment_provider` is written by UPDATE, so `BEFORE INSERT` alone can't cover it) is yours and is preserved with attribution. Two corrections to my review above, one of which is against myself.

**I was wrong about *why* `SECURITY DEFINER` is needed.** I said `revenue_splits` is admin-only under RLS, repeating what your description and the checkout route's comment say. It isn't — the policy's predicate doesn't match its name:

```
 Admins can view own revenue split | SELECT | tenant_id = get_tenant_id() OR is_super_admin()
```

Every member of the tenant, students included, can read their school's split. So an INVOKER function would in fact work today, and the info-disclosure nit in point 5 of my review is void — students can already read that value directly. `DEFINER` is still the right call, but for the inverse reason: it stops this trigger's correctness from resting on a permissive policy whose own name contradicts it. Tighten that policy someday and an INVOKER trigger silently stamps the 80 fallback onto every checkout.

**Point 1 turned out to be demonstrable, not theoretical.** With this branch applied to a local DB, on a tenant whose real split is 70:

```
INSERT tamper -> snapshot = 100  (real split is 70)
UPDATE tamper -> snapshot = 100  (was 70)
```

Both survive, because "an explicit app-layer value always wins" is evaluated before anything else — and on this table the explicit value can come from a student, since the RLS policies restrict rows rather than columns. #527 inverts that rule: computed on INSERT, frozen on UPDATE. The wider version of that hole (a self-inserted `status: 'successful'` row that self-enrols *and* fabricates payout liability) is out of scope for #512 and is now #528.

Point 2 (the UPDATE branch acting as a lazy backfill of legacy rows) and point 3 (the snippet not being safe against a populated database) are addressed there too, along with the 7-case verification run you asked for — it passes, and your case 4 is in it unchanged, since it was the right test to have written.

I'd suggest closing this as superseded by #527 once that lands, unless you'd rather take the changes onto this branch — happy either way, and thanks for a genuinely well-argued PR.

guillermoscript:
@DPS0340 I've pushed the review changes onto this branch directly (via "Allow edits by maintainers") rather than asking you to do another round — it was faster than describing them, and your commit stays as the base with your authorship intact. The follow-up I opened separately is closed as superseded; **this PR is the one to merge.** Push back on any of it, it's your PR.

The commit is `626217a8`. Below is what changed and why, since the description above now predates it.

### 1. The snapshot column is client-writable, so "an explicit value always wins" was the wrong default

`transactions` grants `ALL` to `authenticated` (`20260126190500_lms_complete.sql:3692`) and its RLS policies restrict which **rows** a user may write, never which **columns** (`20260313025318_rls_transactions.sql`). So `school_percentage_snapshot` can be supplied by a student, and it flows straight into `computeOwedBalances()`'s `grossOwed`. On a local DB, against this branch as it stood, with the tenant's real split at 70:

```
INSERT tamper -> snapshot = 100  (real split is 70)
UPDATE tamper -> snapshot = 100  (was 70)
```

Both survived, because the non-NULL early return runs before anything else. Same two statements now:

```
INSERT tamper -> snapshot = 70.00  (real split is 70)
UPDATE tamper -> snapshot = 70.00  (was 70)
```

The column is now **database-owned** rather than database-defaulted: computed from `revenue_splits` on INSERT with no `WHEN` guard (the point is to override what the caller sent), and frozen on UPDATE — the previous value is restored verbatim rather than raising, so a tampering UPDATE is neutralised instead of failing the webhook that happened to carry it.

### 2. The UPDATE branch was a lazy, non-deterministic backfill

Once the trigger is installed, every new row is snapshotted at INSERT — so the UPDATE branch could only ever fire on **legacy NULL rows**, stamping *today's* split on them at whatever arbitrary moment something incidental first touched the row. A May sale refunded next week would be frozen at next week's split. Your description argues against backfilling for exactly that reason, so this was the implementation not matching its own stated intent rather than a disagreement about the goal.

A legacy NULL is now filled in one case only: `payment_provider` is changing, i.e. a provider is activating the row right now. That needs two triggers rather than one, because `OLD` can't be referenced in a `WHEN` clause on a combined `INSERT OR UPDATE` trigger.

### 3. Correction, against my own review: `revenue_splits` is not super-admin-only

I said it was, repeating what your description and the checkout route's comment say. Verified on a local DB, the policy's predicate doesn't match its name:

```
 Admins can view own revenue split | SELECT | tenant_id = get_tenant_id() OR is_super_admin()
```

Every tenant member — students included — can read their school's split. So an INVOKER function would work today, and the info-disclosure nit in point 5 of my review is void. `SECURITY DEFINER` is kept for the inverse reason: it stops this trigger's correctness from resting on a permissive policy whose own name contradicts it. Tighten that policy someday and an INVOKER trigger silently stamps the 80 fallback onto every checkout, for every tenant. The misleading comment in `app/api/payments/checkout/route.ts` is corrected too.

### Also

- The verification snippet is relabelled **LOCAL ONLY** and grown 5 → 7 cases. Your case 4 is in it unchanged — it was the right test to have written. The rollback protects the *data*, not availability: `ALTER TABLE ... DISABLE TRIGGER` locks the live payments table, needs table ownership, and leaves the trigger disabled for other sessions if the script dies with the transaction open.
- The unit tests are relabelled as static drift guards. The `DEFAULT_SCHOOL_PERCENTAGE` one earns its place; the tautological `not.toMatch(/UPDATE\s+transactions\s+SET/i)` is dropped and a guard that the INSERT trigger stays unconditional replaces it. A green run of that file still says nothing about whether the trigger works — the snippet is the coverage.
- The rollback script drops both triggers (and the old single-trigger name) and now says that rolling back re-opens the client-writable column, not just the #496 fallback.

### The verification you asked for, run

```
PASS 1: insert without snapshot filled from revenue_splits (70)
PASS 2: client-supplied snapshot (100) overridden with the real split (70)
PASS 3: missing revenue_splits row falls back to 80
PASS 4: provider activation filled the snapshot (70) — the BEFORE INSERT gap
PASS 5: existing snapshot untouched by a later update (still 70, split now 90)
PASS 6: update tampering with an existing snapshot reverted to 70
PASS 7: legacy NULL row still NULL after an incidental update and a tamper attempt
```

Also on the local DB: the rollback script drops both triggers and the function cleanly, after which the migration re-applies and all 7 pass again; and `SECURITY DEFINER` works from the real checkout shape (`SET LOCAL ROLE authenticated` with a tenant claim, inserting with no snapshot → the tenant's real split, not the 80 fallback).

Suite level: `npm run test:unit` 321 passed (28 files) · `typecheck` clean · `eslint` clean on changed files.

**Still open before merge:** the migration has not been replayed from scratch (`npm run db:reset`) — it was applied by hand onto an existing local schema, so ordering is verified by inspection only. And `supabase db push` to cloud is still to run. Both are mine to do, not yours.

Wider than this issue and now filed separately as #528: the same RLS gap lets a user self-insert a `status: 'successful'` row, which both self-enrols them via `trigger_manage_transactions` and fabricates payout liability.

Thanks for this one — the UPDATE finding was the part that mattered, and it wasn't in the issue.

guillermoscript:
Final check against the issue rather than against the diff — every transaction-insert path in the repo, driven in its own shape on a local DB where the fixture tenant's split is 70:

```
PASS site 1  (public)/checkout/actions.ts:54          mock product
PASS site 2  (public)/checkout/actions.ts:102         mock plan
PASS site 3  payment-requests.ts:359                  manual/offline
PASS site 4  create-payment-intent/route.ts:141       stripe connect
PASS site 5  payments/checkout/route.ts:234           app-layer write agrees
PASS site 6  grant_free_subscription.sql:74           free grant
PASS site 7  create_transaction_for_renewal()         tenant_id defaulted -> snapshot 80.00
PASS risk    unsnapshotted insert + webhook activation -> 70, not the current-split fallback
```

Sites 1–4 are the four the issue lists as unsnapshotted. Two more turned up that it doesn't mention, both SQL-side: `grant_free_subscription()` (sets `tenant_id`, so it snapshots normally) and `create_transaction_for_renewal()`, which omits `tenant_id` entirely — but `transactions.tenant_id` carries a column DEFAULT of the default tenant, so those rows are still tenant-scoped and still get a snapshot. There is no insert path left that produces a NULL.

The last line is the issue's stated risk end to end: a row created by a path that never sets the snapshot, then turned platform-settled by the webhook activation path — the case a `BEFORE INSERT` trigger alone would have missed.

One deliberate deviation from the issue's wording, worth being explicit about since it's the acceptance criterion: it asks for the app-layer write to be kept "as a harmless override." It isn't an override any more — the database overrides *it*. That's the point of the change, since the same column is writable by any authenticated client. `app/api/payments/checkout/route.ts` still computes and sends the value (it agrees, and it keeps the rollback migration a safe lever), but it is no longer authoritative.

Full snippet also re-run after a fresh apply: 7/7 pass.

PR #526, opened 2026-07-25
fix(billing): gate past-due fields on past-due reasons, flatten at-risk fetch (#514)

### LINKED ISSUE #514: Billing-health dashboard misses past-due subscriptions and scheduled-cutoff tenants (#495 follow-up)

guillermoscript:
Closes #514. Builds directly on @DPS0340's #523 — their commit is the first commit on this branch, unmodified, with authorship preserved. This PR is #523 plus the follow-ups from its review.

## What #523 got right (kept as-is)

The union fetch is the correct answer to §1 and §2: three reads merged by tenant id (`tenants.billing_status='past_due'`, tenants owning a `past_due` subscription, `tenants.access_cutoff_at IS NOT NULL`), each contributing an `AtRiskReason`, because PostgREST cannot express a disjunction spanning two tables. Keeping `reasons` as a set rather than collapsing to one enum is what makes the drift visible. The deterministic subscription ranking for §3 is also kept, along with its honest framing: `platform_subscriptions` carries `UNIQUE (tenant_id)` (`20260217040000_platform_billing.sql:46`) and every writer upserts on it, so §3 is unreachable today — the ranking is a trap-removal, not a live fix.

### 1. Past-due-derived fields are gated on the tenant being past due

Widening the population admits tenants whose billing is healthy, but `computeBillingHealth` still derived `pastDueSince`, `graceEndsAt`, `daysUntilDowngrade` and `isEstimate` from `payment_method` alone. That is wrong for the single most common cutoff-only population — a school that already lapsed to free and is now over the free plan's limits:

- `downgradeTenantToFree()` sets `status: 'canceled'` but **never clears `grace_period_end`** (`lib/billing/downgrade-tenant.ts:34-41`); only `confirmManualPayment` and the non-`liveCycle` branch of `changePlan` do.
- The same function then calls `reconcileAccessCutoff()` (`:64`) — which is exactly what stamps the `access_cutoff_at` the new query selects on.

So that tenant arrived with `payment_method: 'manual_transfer'` and a months-old grace stamp, and rendered **"Past due since Jun 1, 2026"** with a red **"Overdue"** in _Downgrades in_. Worse, the sort is ascending on `daysUntilDowngrade`, so its `-46` put it **above every school that still had a downgrade to avert**. A healthy Stripe tenant with a cutoff had the milder version: a *future* renewal date under "Past due since", plus the "Stripe-managed" dunning badge.

Those four fields now require `tenant_past_due` or `subscription_past_due` among the row's reasons. `isEstimate` also reads correctly again: false for a cutoff-only row, which has no downgrade to estimate rather than an unknowable one.

### 2. The no-countdown tail is ordered

Rows without a countdown used to compare equal to each other. They now sort by soonest `accessCutoffAt`, then by name, so the tail is meaningful and the render is stable across reloads.

### 3. Fetch flattened from four sequential round-trips to two

`await verifySuperAdmin()` → `Promise.all([3 reads])` → `tenants.in(missingIds)` → `subscriptions.in(ids)` was two more serial layers than the union needs. The past-due subscription read now takes the full column set instead of just `tenant_id`, so the subscription rows for subscription-only tenants are already in hand — which makes the follow-up `tenants` fetch independent of the subscription fetch, and the two go out together. Feeding both subscription lists to `computeBillingHealth` is safe precisely because of #523's ranking change: duplicate identical rows resolve to the incumbent instead of last-wins.

The super-admin check deliberately stays **ahead** of the reads and is not folded into the `Promise.all` — there is a comment saying so, because it is the kind of thing a later "parallelize everything" pass would break. A non-super-admin caller must not cause tenant billing data to be read at all.

### 4. Testability and typing

- `mergeAtRiskTenants()` is now a pure exported function, so the union — the part of #514 with the actual new logic, and the only part #523 left untested — is unit-testable without a database.
- Row mapping goes through typed helpers instead of `as TenantRow[]`, so a drifting `select` list fails `tsc` instead of rendering blanks.

### 5. UI

- A cutoff already in the past is labelled **"Access paused"** instead of "Cutoff scheduled" — access has already stopped, and `accessCutoffActive` is computed in the pure layer against the injected `now` so it stays deterministic.
- The four metric-card counts are computed in one pass over the list instead of five chained `.filter()` calls.

## Deliberately not in scope

- **i18n.** All strings are hardcoded English, which matches the convention: none of the 9 pages under `app/[locale]/platform/` use `next-intl`, since this console is super-admin-only. Localizing it is a separate change covering all 9 pages.
- **Reconciling `tenants.billing_status` against `platform_subscriptions.status` at write time.** #523 called this out and was right to leave it: this dashboard now stops *hiding* the drift; removing the drift is its own issue.
- No migration, no schema change, no new dependency.

## Testing

- [x] `npx vitest run` — **332 passed (27 files)**, up from 323. `tests/unit/billing-health.test.ts` goes 11 → 20: the stale-grace cutoff-only row reports no countdown and no "past due since", a live past-due school outranks an already-downgraded one, the no-countdown tail ordering, `accessCutoffActive` on both sides of `now`, and five cases for `mergeAtRiskTenants` (union without duplicates, reason accumulation, subscription-only surfacing, a subscription whose tenant row is missing, and the cutoff read not clobbering a past-due row).
- [x] `npx tsc --noEmit` — clean.
- [x] `npx eslint` on all four changed files — clean.
- [ ] **Not verified in a browser.** Docker is not running on this machine and there are no local Supabase containers, so the page was never rendered against real rows. Everything above is unit-level.

### QA script for the reviewer

With a local stack up (`supabase start && npm run db:reset`), as a super admin at `/platform/billing-health`:

1. **Cutoff-only, already downgraded** — pick a tenant, set `platform_subscriptions` to `status='canceled'`, `payment_method='manual_transfer'`, `grace_period_end` a month in the past, and `tenants.access_cutoff_at` a week in the future, with `billing_status` *not* `past_due`.
   Expect: listed with a single **Cutoff scheduled** badge, `—` in both _Past due since_ and _Downgrades in_, and positioned **below** any school with a live countdown. On master it is absent entirely; on #523 alone it shows a red "Overdue" at the top of the table.
2. **Cutoff already in force** — same tenant, `access_cutoff_at` in the past. Badge should read **Access paused**.
3. **Subscription-only past due** — set `platform_subscriptions.status='past_due'` while leaving `tenants.billing_status='active'`. Expect the school listed with a **Subscription past due** badge (invisible on master).
4. **Regression** — a genuinely past-due manual-transfer school with a future `grace_period_end` still shows its countdown, still colour-codes at ≤3 and ≤7 days, and still sorts first.
5. Metric cards: the first three count past-due schools only; the fourth counts scheduled cutoffs. Adding an over-limit school must not move the first three.

https://claude.ai/code/session_01WJDnXnWmdBDTX33bvTx62K

### DIFF
--- a/app/[locale]/platform/billing-health/page.tsx
+++ b/app/[locale]/platform/billing-health/page.tsx
@@ -1,4 +1,5 @@
 import { getAtRiskTenants } from '@/app/actions/platform/billing-health'
+import type { AtRiskReason } from '@/lib/billing/billing-health'
 import { Card, CardContent, CardHeader, CardTitle } from '@/components/ui/card'
 import { Badge } from '@/components/ui/badge'
 import { format } from 'date-fns'
@@ -6,19 +7,33 @@ import {
   IconAlertTriangle,
   IconClockPause,
   IconCreditCardOff,
+  IconLockExclamation,
 } from '@tabler/icons-react'
+
+const REASON_LABELS: Record<AtRiskReason, string> = {
+  tenant_past_due: 'Past due',
+  subscription_past_due: 'Subscription past due',
+  access_cutoff_scheduled: 'Cutoff scheduled',
+}
 
 export default async function PlatformBillingHealthPage() {
   const atRisk = await getAtRiskTenants()
 
-  const manualTransfer = atRisk.filter((t) => t.paymentMethod === 'manual_transfer')
-  const stripeDunning = atRisk.filter((t) => t.paymentMethod !== 'manual_transfer')
+  // The metric cards deliberately keep counting past-due tenants only, so the
+  // numbers do not silently change meaning now that #514 also lists tenants
+  // that are merely over their plan limits. Cutoffs get their own card.
+  const pastDue = atRisk.filter(
+    (t) => t.reasons.includes('tenant_past_due') || t.reasons.includes('subscription_past_due'),
+  )
+  const manualTransfer = pastDue.filter((t) => t.paymentMethod === 'manual_transfer')
+  const stripeDunning = pastDue.filter((t) => t.paymentMethod !== 'manual_transfer')
+  const cutoffScheduled = atRisk.filter((t) => t.reasons.includes('access_cutoff_scheduled'))
   const urgent = manualTransfer.filter((t) => t.daysUntilDowngrade !== null && t.daysUntilDowngrade <= 3)
 
   const metricCards = [
     {
       title: 'Past-Due Schools',
-      value: String(atRisk.length),
+      value: String(pastDue.length),
       sub: 'Total schools currently behind on payment',
       icon: IconAlertTriangle,
       bg: 'bg-red-50 dark:bg-red-950/40',
@@ -40,6 +55,14 @@ export default async function PlatformBillingHealthPage() {
       bg: 'bg-blue-50 dark:bg-blue-950/40',
       iconColor: 'text-blue-600 dark:text-blue-400',
     },
+    {
+      title: 'Access Cutoff Scheduled',
+      value: String(cutoffScheduled.length),
+      sub: 'Over plan limits — access pauses on the cutoff date',
+      icon: IconLockExclamation,
+      bg: 'bg-purple-50 dark:bg-purple-950/40',
+      iconColor: 'text-purple-600 dark:text-purple-400',
+    },
   ]
 
   return (
@@ -47,11 +70,12 @@ export default async function PlatformBillingHealthPage() {
       <div className="mb-8">
         <h1 className="text-2xl font-bold tracking-tight">Billing Health</h1>
         <p className="text-sm text-muted-foreground mt-1">
-          Schools currently past-due, with their grace-period countdown to automatic downgrade.
+          Schools that are past-due or over their plan limits, with their countdown to automatic
+          downgrade or access cutoff.
         </p>
       </div>
 
-      <div className="mb-8 grid gap-3 sm:grid-cols-3" data-testid="billing-health-metrics">
+      <div className="mb-8 grid gap-3 sm:grid-cols-2 lg:grid-cols-4" data-testid="billing-health-metrics">
         {metricCards.map((card) => (
           <Card key={card.title} className="relative overflow-hidden">
             <CardContent className="p-5">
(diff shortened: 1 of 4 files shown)


guillermoscript:
Superseded — the follow-ups landed directly on #523 as `bdc9640f`, on top of @DPS0340's commit, so there is one PR for #514 instead of two competing ones. Nothing here is lost: the commit content is identical, and the before/after captures are posted on #523.

Closing the branch as well.

Issue #540, opened 2026-07-25
EPIC: Correctness pass over the three shipped epics — #458, #388, #493


guillermoscript:
# EPIC: Correctness pass over the three shipped epics — #458, #388, #493

## Summary

Epics #458 (plan-change & subscription lifecycle), #388 (evidence-based learning engine) and #493 (payout & non-payment enforcement) are all delivered — 27 of their 30 sub-issues are merged and the unit suite is green at **370/370**.

This epic is the audit of what those green checkmarks do **not** cover. Five independent read-only sweeps (2026-07-25, one per subsystem) produced 43 findings; 38 survived a vetting pass in which every headline claim was re-read in the source and, where it was a database claim, executed against **both the local Postgres and the cloud project**.

The pattern across all three epics is the same and it is worth stating once, because it explains why so much shipped green:

> **The tests pin the modules; nothing tests the callers.** `platform-plan-change.test.ts` passes while the route that calls it cannot run. `solana-pull-decision.test.ts` passes on a fixture (`cancel_at: null`) the schema forbids. `transaction-split-snapshot-backstop.test.ts` greps migration text and says so in its own header. `payouts-owed.test.ts` uses only round numbers. And `mcp-server/` — which holds the FSRS scheduler, the interleaving gate and the Elo selector — has **no test runner configured at all**.

Three of the findings are live right now on the cloud project and were confirmed there by query, not inference.

**Deployment note for everyone working on this epic:** "production" here means the cloud Supabase project only. There is no deployed app and no real users. **Breaking changes are fine** — do not write backfill dances, compatibility shims or phased rollouts for existing rows. Drop and recreate. Prefer the correct end state over the migration-safe one.

---

### 1.1 The #538 INSERT lockdown was never applied to cloud 🔴 `security`

```
select has_table_privilege('authenticated','public.transactions','INSERT');  -- true
```

[...]

Issue #538, opened 2026-07-25
Solana settlement columns are caller-controlled at INSERT — a pending row can under-quote what it owes (#528 follow-up)


guillermoscript:
Found while implementing #528 (PR #537), 2026-07-25. Verified against a local database.

## Problem

#528 closed the two impacts that depended on an untrusted caller producing a `successful` transaction: the INSERT policy on `transactions` now pins `status = 'pending'`, and table-level `UPDATE` is revoked from `authenticated` in favour of a three-column grant.

That invariant is "an untrusted caller may open a transaction, but may not declare it paid". It deliberately says nothing about **what a pending row claims it is owed** — and the Solana settlement columns are exactly that:

```
settlement_currency  text     -- 'sol' | 'usdc'
settlement_base      bigint   -- amount in lamports / USDC base units
settlement_mint      text
settlement_sol_usd   numeric(20,8)
```

All four are still writable by `authenticated` on INSERT, because `app/api/payments/checkout/route.ts:233` legitimately writes them on the user-scoped client and the fix in #528 did not narrow INSERT columns.

## Impact

The on-chain payment is verified against the row's **own** `settlement_base`, in `lib/payments/solana-reconcile.ts:103-106`:

```ts
if (tx.settlement_base != null && tx.settlement_currency) {
  vTotalBase = Number(tx.settlement_base)
  vSplTokenStr = tx.settlement_mint
  vDecimals = tx.settlement_currency === 'usdc' ? 6 : 9
}
```

`vTotalBase` becomes the `totalBase` handed to `verifySplitTransfer`. Both callers reach it — `app/api/payments/solana/verify/route.ts:27` (`reconcileSolanaOneTimeTransaction`) and the `/api/cron/solana-reconcile` backstop. A student can therefore skip the checkout route and `POST /rest/v1/transactions` directly:

```json
{
  "user_id": "<self>", "tenant_id": "<own tenant>",
  "product_id": <a $100 course>, "amount": 100, "currency": "usd",
  "status": "pending", "payment_provider": "solana",
  "settlement_currency": "sol", "settlement_base": 1
}
```

The row is `pending`, so it passes the #528 policy cleanly. They then pay **1 lamport**, the verify endpoint compares the on-chain transfer against `settlement_base: 1`, finds it satisfied, and flips the row to `successful` on the service-role client — which fires `trigger_manage_transactions` and grants the entitlement. Full course access for a fraction of a cent, through the legitimate payment path.

[...]

Issue #543, opened 2026-07-25
Lock the write side: grades, progress and practice rows are self-issuable


guillermoscript:
Part of EPIC #540. §1.3.

## What's wrong

Four write-side policies constrain **who the row belongs to** but not **whether the writer earned it**. All four verified live in `pg_policies`.

### `lesson_completions` — completion is self-issuable, and certificates count it

```
INSERT WITH CHECK: (SELECT auth.uid()) = user_id
```

That is the entire policy (`supabase/migrations/20260131170137_fix_lesson_completions_rls.sql:13-16`). No course gate, no tenant, no check that the lesson exists or that the caller can reach it. Any authenticated user can insert a completion for **any** `lesson_id`.

That matters because `app/api/certificates/issue/route.ts` — student self-serve branch, `:16-85` — checks auth, tenant ownership of the course and duplicates, but **never** calls `hasCourseAccess` / `resolveCourseAccessState`. Eligibility falls through to a raw `lesson_completions` count at `:176-186`. So the school's credential can be minted by someone who never held access.

### `exam_submissions` — no access gate on submit

```
INSERT WITH CHECK: ((SELECT auth.uid()) = student_id) AND (tenant_id = (SELECT get_tenant_id()))
```

(`supabase/migrations/20260313151603_rls_exams_exam_submissions.sql:46-48`.) No `has_course_access`. A student with no entitlement, or one past their school's `access_cutoff_at`, can POST submissions that then enter the AI grading pipeline (`app/actions/exam-grading.ts`) on the school's plan budget. Paired with the sibling issue in this epic (§1.2) they can read the questions first.

### `lesson_checkpoint_attempts` — the grade is client-supplied

The policy pins `user_id`, checkpoint/lesson/exercise consistency, `c.is_enabled` and `has_course_access` (that is #532/#535's fix and it holds). But `score`, `passed`, `completed`, `evaluation`, `response`, `evaluator_type` and `attempt_number` are **all unconstrained**, and `authenticated` holds the INSERT grant.

`app/api/lesson-checkpoints/[checkpointId]/attempt/route.ts:285-299` grades server-side and then inserts **with the caller's own RLS client** — the code comment says "so DB policies are the last word", but the policies say nothing about the graded columns. So `{passed: true, score: 100, completed: true}` satisfies any `is_required` checkpoint without answering; `lib/checkpoints/load.ts:73-96` derives `latestAttempt.passed` straight from the newest row, and required checkpoints gate lesson progression.

[...]

Issue #542, opened 2026-07-25
exam_questions and question_options are readable by every authenticated user in every tenant


guillermoscript:
Part of EPIC #540. §1.2.

## What's wrong

Three SELECT policies are `USING (true)`. Verified identical on **cloud and local**:

```sql
select tablename, policyname, qual from pg_policies
 where tablename in ('exam_questions','question_options','lesson_comments') and cmd='SELECT';
-- exam_questions   | Authenticated users can view exam questions   | true
-- question_options | Authenticated users can view question options | true
-- lesson_comments  | Authenticated users can view lesson comments  | true
```

`exam_questions` and `question_options` are the only two policies in the audited set with **no tenant predicate at all**.

- `exam_questions` carries `correct_answer`, `grading_rubric`, `ai_grading_criteria`, `expected_keywords`.
- `question_options` carries `is_correct`.

So any authenticated user — of **any** tenant, holding no entitlement, past their school's `access_cutoff_at` — can read every exam's answer key and every AI grading rubric in the database directly over PostgREST. Both the #509 exam gate and the #494 access cutoff are bypassable for the thing exams exist to protect.

Origin: `supabase/migrations/20260313152048_rls_exam_child_tables.sql:8-10,22-24` created both as `USING (true)`. #509 (`supabase/migrations/20260724150000_content_entitlement_rls.sql:96-110`) replaced the **`exams`** SELECT policy with a proper tenant + staff/author + `has_course_access` gate and stopped at the parent table. `exam_answers` was fixed separately in `20260704200000_fix_advisor_security_findings.sql`. Questions and options were left behind.

`lesson_comments` is the same shape with lower stakes: every authenticated user reads every lesson discussion thread in every tenant.

## Also in scope — the preview-lesson overshoot

`supabase/migrations/20260724150000_content_entitlement_rls.sql:57-59,68` gives authenticated users a preview branch of `(is_preview = true AND status = 'published')`, justified as "a logged-in prospective buyer must not see less than an anonymous one". The anon policy it mirrors carries one more condition (confirmed live): an `EXISTS` requiring the **parent course** to be `status = 'published'`.

So a teacher building a course in draft who marks a lesson `is_preview` + `published` exposes that lesson body to every signed-in member of the tenant, students included, before the course is ever published — while anonymous visitors correctly cannot see it. The parity argument is one-directional; the implementation overshoots it. Content leak introduced *by* the hardening migration.

## Scope

**In scope** — one migration replacing four policies:

[...]

PR #552, opened 2026-07-25
fix(security): apply the transactions INSERT lockdown to cloud and make migration drift detectable (#541)

### LINKED ISSUE #528: transactions RLS restricts rows but not columns — a student can self-insert a 'successful' sale

### LINKED ISSUE #541: Apply the transactions INSERT lockdown to cloud and make migration drift detectable
Part of EPIC #540. §1.1.

## What's wrong

The `transactions` INSERT lockdown shipped in #538 / PR #539 is **not applied to the cloud project**. Verified by query against cloud:

```sql
select has_table_privilege('authenticated','public.transactions','INSERT');  -- true
select policyname, with_check from pg_policies
 where tablename='transactions' and cmd='INSERT';
-- "Users can create own transactions"
-- ((SELECT auth.uid()) = user_id) AND (tenant_id = (SELECT get_tenant_id())) AND (status = 'pending'::transaction_status)
```

That is the **#528** policy shape. `supabase/migrations/20260725180000_transactions_insert_lockdown.sql` does two things that are both absent on cloud:

- `:116` — `REVOKE INSERT ON TABLE public.transactions FROM authenticated, anon`
- `:150-158` — pins `settlement_base`, `settlement_currency`, `settlement_rate`, `settlement_quoted_at` to `IS NULL` on the INSERT policy

So the scenario #538 was filed for is open in production. A caller opens a pending row quoting `settlement_base: 1` against a $49 product, pays that amount on-chain, and `/api/payments/solana/verify` — which verifies the transfer against **the row's own `settlement_base`** (`lib/payments/solana-reconcile.ts:104`) — flips it to `successful` on the service-role client and grants the entitlement. The same grant leaves `amount` caller-controlled, and `getPayoutsOwed` sums `amount` into real payout liability.

PR #539 already moved every legitimate insert to `createAdminClient()`, so **nothing depends on the grant**. It is pure attack surface.

## Root cause — why it got lost

Cloud migration stamps have drifted from the repo filenames. Cloud `supabase_migrations.schema_migrations` holds:

```
20260724120000, 20260724130000, 20260724140000, 20260724150000, 20260725100000,
20260725192946, 20260725213441, 20260725213508, 20260725213610
```

The repo holds `20260725110000_transaction_split_snapshot_backstop`, `20260725160000_entitlement_gated_enrollment_inserts`, `20260725170000_transactions_column_hardening`, `20260725180000_transactions_insert_lockdown`. The content of three of those was applied ad hoc via the MCP `apply_migration` tool, which stamps a **fresh** timestamp, instead of via `supabase db push`. The fourth was never applied at all.

Consequence: `schema_migrations` cannot answer "what is missing from cloud". Comparing the local `supabase/migrations/` listing against it produces false positives for the re-stamped ones and hides the genuinely-missing one. That is exactly how `20260725180000` was lost — and #500 and the epic's §3.1 note record the same class of gap twice before.

## Scope

**In scope**
- Apply `supabase/migrations/20260725180000_transactions_insert_lockdown.sql` to cloud.
- Reconcile the four drifted stamps so `schema_migrations` reflects reality (repair the ledger; do not re-run already-applied DDL).
- Add a check that fails loudly on drift — a `scripts/verify-cloud-schema.ts` (or a `supabase/snippets/` SQL file wired into a documented step) asserting at minimum:
  - `has_table_privilege('authenticated','public.transactions','INSERT')` is `false`
  - the INSERT policy's `with_check` contains all four `settlement_*` NULL pins
  - the `authenticated` UPDATE column grant is exactly `(status, provider_subscription_id, stripe_payment_intent_id)`
  - `before_transaction_split_snapshot_insert` / `_update` triggers exist and are enabled
- Document the rule in `docs/MIGRATIONS.md`: cloud changes go through `supabase db push`, never through ad-hoc `apply_migration`, precisely because the latter breaks drift detection.

**Out of scope**
- Any change to the migration's SQL. It is correct; it was simply never run.
- The `20260725110000` / `160000` / `170000` content — already live on cloud under different stamps, confirmed by querying the grants, policies and `pg_trigger`.

## Acceptance criteria

- [ ] `select has_table_privilege('authenticated','public.transactions','INSERT')` returns **false** on cloud.
- [ ] Cloud's INSERT policy `with_check` includes the four `settlement_* IS NULL` pins.
- [ ] The drift check exists, is runnable in one command, and **fails** if you manually re-grant INSERT (prove it by trying, then reverting).
- [ ] `npm run test:unit` stays green (370/370 baseline at `146a8cfa`).
- [ ] A dry run of `supabase db push` against cloud reports **no** pending migrations.
- [ ] `docs/MIGRATIONS.md` states the push-only rule and the reason.

## Notes

There are no real users and no deployed app — the cloud Supabase project is the whole of "production". Breaking changes are fine; if repairing the ledger is easier by resetting and re-pushing, do that.

`tests/unit/transaction-split-snapshot-backstop.test.ts:1-17` says in its own header that it greps migration text and that "a green run of this file means nothing about whether the migration was ever applied". That is the gap this issue closes — the new check must query the live database, not the repo.

**Effort:** S · **Risk:** LOW/MED (every insert path is service-role or SECURITY DEFINER; a user-scoped insert missed in the #538 inventory would start failing loudly with `permission denied for table transactions` — that is the intended signal, not a regression)

guillermoscript:
Closes #541. Part of EPIC #540, §1.1.

## What was wrong

`20260725180000_transactions_insert_lockdown.sql` (#538) was **never applied to cloud**. Verified by querying the live catalog before touching anything:

```
has_table_privilege('authenticated','public.transactions','INSERT')  ->  true
INSERT policy with_check  ->  the #528 shape, no settlement_* pins
```

So the #538 scenario was open on the only production database this project has: open a pending row quoting `settlement_base: 1` against a priced product, pay 1 lamport, and `/api/payments/solana/verify` — which verifies the transfer against **the row's own `settlement_base`** (`lib/payments/solana-reconcile.ts:104`) — flips it to `successful` on the service-role client and grants the entitlement.

## What this does

**1. Applied the lockdown to cloud.** `REVOKE INSERT ... FROM authenticated, anon` plus the INSERT policy re-created with all four `settlement_* IS NULL` pins. Confirmed by query afterwards:

| assertion | before | after |
|---|---|---|
| `authenticated` INSERT | `true` | **`false`** |
| `anon` INSERT | `false` | `false` |
| `service_role` INSERT | `true` | `true` (untouched — every webhook/reconciler/cron depends on it) |
| INSERT policy settlement pins | none | **all four present** |

Nothing depended on the grant — PR #539 had already moved both user-scoped inserts onto `createAdminClient()`. The two user-scoped `.update()` calls remaining in `app/api/payments/checkout/route.ts` only touch `provider_subscription_id` and `status`, both inside the #528 three-column UPDATE grant, so they still work.

**2. Repaired the drifted ledger** (`20260726005843_repair_migration_ledger_541.sql`). Cloud now matches the repo exactly: **171 files, 171 stamps, zero pending, zero orphans.**

**3. `npm run verify:cloud`** — a drift check that queries the live database and exits non-zero on drift.

**4. `docs/MIGRATIONS.md`** now leads with the push-only rule, the reason, and the incident table.

## Root cause

Cloud stamps had drifted from repo filenames — four migrations applied via MCP `apply_migration`, which stamps a **fresh** timestamp:

| Repo file | Cloud stamp before | now |
|---|---|---|
| `20260721120000_add_binance_personal_provider` | `20260725213441` | ✅ |
| `20260725110000_transaction_split_snapshot_backstop` | `20260725213508` | ✅ |
| `20260725160000_entitlement_gated_enrollment_inserts` | `20260725213610` | ✅ |
| `20260725170000_transactions_column_hardening` | `20260725192946` | ✅ |
| `20260725180000_transactions_insert_lockdown` | **never applied** | ✅ |

All five read as "pending" against the ledger, so the genuinely missing one hid among four false positives. Note the issue named three drifted stamps; there were **four** — `add_binance_personal_provider` had drifted too. Each of the four was confirmed genuinely applied by querying what its DDL created (the `binance_personal` CHECK value, `has_course_access` in the enrollments INSERT policy, both snapshot triggers, the 3-column UPDATE grant) before being treated as a ledger-only repair, so the repair re-runs no DDL.

The repair migration is committed under the stamp `apply_migration` assigned it, so it is not itself an orphan. It is idempotent — on a fresh `supabase db reset` the four UPDATEs match nothing and the INSERT no-ops.

## `npm run verify:cloud`

```
SUPABASE_ACCESS_TOKEN=sbp_... npm run verify:cloud
```

Against cloud right now — **10/10 PASS**:

```
PASS  every repo migration is applied to cloud                          all 171 applied
PASS  cloud carries no migration stamp that matches no repo file        ledger matches repo filenames exactly
PASS  authenticated holds no INSERT on transactions (#538)              revoked
PASS  anon holds no INSERT on transactions (#538)                       revoked
PASS  no column-level INSERT grant to authenticated survives            (none)
PASS  service_role retains INSERT                                       granted
PASS  authenticated UPDATE grant is exactly (...)                       matches
PASS  INSERT policy pins status + all four settlement columns (#538)    all pins present
PASS  trigger before_transaction_split_snapshot_insert ... enabled      enabled
PASS  trigger before_transaction_split_snapshot_update ... enabled      enabled
```

It asserts **ledger integrity** and the **#512/#528/#538 payment invariants** against catalog state rather than migration text — so a later migration re-widening things, or a schema dump re-applying the original `GRANT ALL ON TABLE transactions TO authenticated`, fails it too.

It uses the Management API over HTTPS rather than a Postgres connection because the pooler on port 5432 is unreachable from this network (`supabase migration list` → `tls error … i/o timeout`), the same reason `docs/MIGRATIONS.md` already documents an HTTPS fallback.

## On the "prove it fails by re-granting" criterion

I did not re-grant INSERT on the live payments table. Re-opening a write grant on production to watch a script go red isn't a trade worth making.

Instead the failure modes are proven as **repeatable tests** in `tests/unit/verify-cloud-schema.test.ts` (13 tests), driven from a healthy fixture that is the *exact* catalog state read back from cloud — including how Postgres renders the policy expression back (`'pending'::transaction_status`), which is the part a hand-written fixture gets wrong. Covered: INSERT re-granted to `authenticated`, to `anon`, a column-level INSERT grant reintroduced, the pre-fix #528 policy shape, the policy dropped, the UPDATE grant widened, `service_role` losing INSERT, a trigger missing or disabled, a migration never applied, an orphan stamp, and the real pre-#541 ledger.

## Known gap

`supabase db push --dry-run` (an acceptance criterion) is **unverified** — the CLI cannot reach the pooler from this network at all. The equivalent was verified by querying `schema_migrations` directly and diffing against `supabase/migrations/`: 171 vs 171, no pending, no orphans. Worth a confirming run from a network with port 5432 open.

## QA

```bash
npm run typecheck            # clean
npm run test:unit            # 383 passed (370 baseline + 13 new)
npx eslint scripts/verify-cloud-schema.ts scripts/lib/ tests/unit/verify-cloud-schema.test.ts   # clean
npm run build                # succeeds
SUPABASE_ACCESS_TOKEN=sbp_... npm run verify:cloud   # 10/10 PASS
```

https://claude.ai/code/session_01HPFyHjqzSR6Ku1gyWjeQie

### DIFF
--- a/docs/MIGRATIONS.md
+++ b/docs/MIGRATIONS.md
@@ -81,6 +81,13 @@ It reads catalog state, not migration text, so it also catches a *later*
 migration re-widening something — including a schema dump re-applying the
 original `GRANT ALL ON TABLE transactions TO authenticated`.
 
+**Run it from `master`, after merging.** The two ledger checks compare cloud
+against the migrations in your *current checkout*. On a feature branch, any
+migration another in-flight branch has already applied to cloud shows up as an
+orphan, and any migration on your branch not yet applied shows up as pending —
+neither is real drift. The payment-invariant checks are branch-independent and
+meaningful anywhere.
+
 The token is a [personal access token](https://supabase.com/dashboard/account/tokens).
 It uses the Management API over HTTPS rather than a Postgres connection because
 port 5432 is blocked on some networks this project is developed from — the same
--- a/docs/MIGRATIONS.md
+++ b/docs/MIGRATIONS.md
@@ -18,6 +18,79 @@ Edit the generated file with your SQL. Always make migrations **idempotent** whe
 
 ## Applying Migrations
 
+### The rule: cloud changes go through `supabase db push`
+
+**Never apply a migration to cloud with the MCP `apply_migration` tool, the SQL
+editor, or a raw Management API call unless you also stamp it with the version
+from its filename.** This is not style preference — it is what keeps drift
+detectable.
+
+`supabase_migrations.schema_migrations` is the only record of what cloud has
+run. `supabase db push` stamps each migration with **the timestamp from its
+filename**. Every ad-hoc path stamps a **fresh** timestamp instead. Once the
+ledger holds a stamp that matches no file in `supabase/migrations/`, the
+question "what is missing from cloud?" stops having a usable answer: the
+re-stamped migrations show up as *pending* (false positives), and a genuinely
+missing migration is indistinguishable from them.
+
+That is not hypothetical. It is how issue #541 happened:
+
+| Repo file | Cloud stamp before #541 |
+|---|---|
+| `20260721120000_add_binance_personal_provider` | `20260725213441` |
+| `20260725110000_transaction_split_snapshot_backstop` | `20260725213508` |
+| `20260725160000_entitlement_gated_enrollment_inserts` | `20260725213610` |
+| `20260725170000_transactions_column_hardening` | `20260725192946` |
+| `20260725180000_transactions_insert_lockdown` | **never applied** |
+
+Four migrations were applied through `apply_migration` and re-stamped. The
+fifth — the `transactions` INSERT lockdown from #538, a `severity:critical`
+payments fix — was never applied at all, and hid among the four false
+positives for a day. Diffing the repo against the ledger reported all five as
+pending, so the real gap looked like more of the same noise.
+
+If you do have to apply SQL out of band, **immediately** repair the stamp:
+
+```bash
+# Preferred — the CLI's purpose-built command for this
+supabase migration repair --status applied <VERSION_FROM_FILENAME>
+supabase migration repair --status reverted <WRONG_FRESH_STAMP>
+```
+
+Then run `npm run verify:cloud` (below) and confirm it is clean before moving on.
+
+### Verifying cloud matches this repo
+
+```bash
+SUPABASE_ACCESS_TOKEN=sbp_... npm run verify:cloud
+```
+
+Queries the **live** database and exits non-zero on drift. It asserts both
+halves of the problem above:
+
+- **Ledger integrity** — every repo migration is stamped on cloud under its own
+  filename, and cloud carries no stamp that matches no file.
+- **Payment invariants** (#512 / #528 / #538) — `authenticated` and `anon` hold
+  no INSERT on `transactions`, no column-level INSERT grant survives,
+  `service_role` still has INSERT, the `authenticated` UPDATE grant is exactly
+  `(status, provider_subscription_id, stripe_payment_intent_id)`, the INSERT
+  policy pins `status` plus all four `settlement_*` columns to NULL, and both
+  `before_transaction_split_snapshot_*` triggers exist and are enabled.
+
+It reads catalog state, not migration text, so it also catches a *later*
+migration re-widening something — including a schema dump re-applying the
+original `GRANT ALL ON TABLE transactions TO authenticated`.
+
+The token is a [personal access token](https://supabase.com/dashboard/account/tokens).
+It uses the Management API over HTTPS rather than a Postgres connection because
+port 5432 is blocked on some networks this project is developed from — the same
+reason Option 2 below exists.
+
+> Unit tests cannot do this job. `tests/unit/verify-cloud-schema.test.ts` and
+> `tests/unit/transaction-split-snapshot-backstop.test.ts` prove the *rules* are
+> right; neither opens a database connection. A green `npm run test:unit` says
+> nothing about whether a migration was ever applied. Only `verify:cloud` does.
+
 ### Option 1: Supabase CLI (preferred)
 
 ```bash
@@ -30,7 +103,10 @@ This connects via the Supabase connection pooler and applies all pending migrati
 
 ### Option 2: Management API (fallback)
 
-When the CLI can't connect, push migrations via the Supabase Management API:
+When the CLI can't connect, push migrations via the Supabase Management API.
+**This is the path that creates drift** — the `INSERT INTO schema_migrations`
+line below is not optional bookkeeping, it is the entire reason this fallback is
+safe to use:
 
 ```bash
 # 1. Get your access token (stored in macOS keychain by the CLI)
@@ -91,16 +167,22 @@ process.stdout.write(JSON.stringify({query: 'SELECT version, name FROM supabase_
 
 ### What's pending locally?
 
-Compare local files against remote records:
-
 ```bash
-# Local migration versions
-ls supabase/migrations/*.sql | sed 's|.*/||' | sed 's/_.*//'
-
-# Remote versions (use one of the methods above)
+SUPABASE_ACCESS_TOKEN=sbp_... npm run verify:cloud
 ```
 
-Any local version not in the remote list needs to be applied.
+Use this rather than eyeballing a diff. The manual comparison —
+
+```bash
+ls supabase/migrations/*.sql | sed 's|.*/||' | sed 's/_.*//'
+```
+
+— is only trustworthy when the ledger is clean. If any migration was ever
+applied out of band, a local version missing from the remote list means *either*
+"never applied" *or* "applied under a different stamp", and nothing in the diff
+tells you which. `verify:cloud` reports the two cases separately (`not on cloud`
+vs `orphan stamps`) and additionally checks that the payment invariants actually
+hold, which no filename comparison can do.
 
 ## Running the Production Seed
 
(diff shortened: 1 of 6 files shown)


guillermoscript:
### Correction to the "10/10 PASS" figure above

That run was accurate when taken, but is now stale, and the reason is worth recording.

Since then PRs #553 (#542) and #554 (#543) applied their own migrations to cloud — `20260726013256`, `20260726015858`, `20260726100000`, `20260726110000`. None of those files exist on this branch, so `verify:cloud` run **from this branch** now reports them as orphan stamps: **9/10 PASS, 1 FAIL**.

That is not drift. The two ledger checks compare cloud against the migrations in the *current checkout*, which makes them branch-relative:

- a migration another in-flight branch has already applied → reads as an **orphan**
- a migration on your branch not yet applied → reads as **pending**

Fixed as guidance rather than logic (a5a921be): the orphan detail line now names the possibility and says to re-check on master, and `docs/MIGRATIONS.md` says to run it from `master` after merging. Filtering by "is this stamp on some other branch" would need a remote ref walk and would silently excuse the exact condition the check exists to catch — so it stays loud.

The eight **payment-invariant** checks are branch-independent and pass everywhere. The security outcome this PR is actually about is unaffected:

```
authenticated INSERT on transactions  ->  revoked (postgres + service_role only)
INSERT policy                          ->  all four settlement_* IS NULL pins present
```

Once #553 and #554 merge, master will hold every file and the ledger reads clean again.

Issue #570, opened 2026-07-27
MCP widget UX polish: i18n, colour semantics, CTAs, unused payload data, card alignment


guillermoscript:
Polish findings from rendering every widget against committed fixtures. None of these are broken output — each one works and reads badly. Independently takeable; do them in any order.

**Repro for all of it**

```bash
cd mcp-server && MCP_DEMO_WIDGETS=1 npm run dev
# inspector: http://localhost:3000/inspector?autoConnect=http%3A%2F%2Flocalhost%3A3000%2Fmcp
./scripts/shoot-demo-widgets.sh dark   # or: light
```

See `mcp-server/docs/WIDGET_DEMO_DATA.md`.

---

### 1. Widgets are English-only while the content is Spanish

Every string the widget itself owns is hardcoded English — `True` / `False` and `Back` / `Next` in `practice-player`, `Save grade`, `Show at risk only`, `last active`, and dates as `Jul 24, 2026`. Rendered next to Spanish course content it reads broken, and en/es is a product requirement, not an option.

`useWidget()` exposes `locale` from the host. At minimum: date/number formatting via `Intl` with that locale, and a small string table per widget.

### 2. Brand purple is used as a mid value on a red→green scale

In `school-overview` and `student-progress-roster`, progress bars run amber → purple → green, so "64% = purple" reads as a status when purple is the *brand* colour, used everywhere else for non-semantic accents. Two meanings, one colour.

Pick one: either status uses a red/amber/green ramp only and brand stays out of it, or the bars are all brand and severity is carried some other way. Related: `school-overview` KPI tiles colour only 2 of 6 numbers with no stated rule, and its "3 published · 2 draft" caption omits the 1 archived course so the parts don't sum to the 6 shown above it.

### 3. exam-readiness buries the thing to act on

Topics render in payload order, so the weakest (12% "Compilador de React") is last, and all six "Practice this" buttons are equally loud. The whole point of the report is *what do I fix first*.

Sort by mastery ascending (or group weak/solid), de-emphasise the buttons on topics already mastered, and show days-until-exam — `exam.exam_date` is in the payload and only rendered as a date.

### 4. my-learning's next lesson isn't a button

The "Next: Lesson 18 · …" pill is grey, passive, and sits in the footer of the card, while the card's actual job is to get the student back into the lesson. Make the primary action look like one.

### 5. course-catalog can't convert

- No price anywhere, on any card.
[...]

PR #575, opened 2026-07-27
fix(widgets): stop rendering missing data as if it were real (#568)

### LINKED ISSUE #568: widgets: missing data rendered as real data (raw user ids, ungraded exams as "—", filter-blaming empty state)
Two widgets present missing data as if it were real data.

**Repro**

```bash
  --tool lms_demo_student_progress_roster variant=default --theme dark --output roster.png
  --tool lms_demo_course_dashboard variant=empty --theme dark --output empty.png
```

## 1. Roster: a student with no profile name renders as a raw UUID

`resources/student-progress-roster/widget.tsx:174`

```tsx
const name = s.student_name ?? s.student_id.slice(0, 8);
```

`profiles.full_name` is nullable, so a real roster shows `u-008` in the name column and `U-` in the avatar circle (`:186`). It reads like corrupted data, and a teacher cannot tell which student it is.

**Fix:** fall back to something human — "Unnamed student", or the email if the tool can get it (note: `profiles` has no `email`; it comes from `auth.admin.getUserById()`, which a widget tool cannot call on the caller's RLS client — so either the server resolves it or the copy stays generic). Avatar initials should degrade to a neutral glyph, not `U-`.

## 2. Roster: "1 exam" next to "—" average

`:226-229` renders `exam_avg == null` as `—` while the line under it says `1 exam`. A student who has submitted an exam that has not been graded yet reads as *took an exam, scored nothing*.

**Fix:** when `exam_count > 0 && exam_avg == null`, say so ("1 exam · not graded") instead of showing an em dash that reads as "no data".

## 3. course-dashboard: the true-empty state blames a filter

`resources/course-dashboard/widget.tsx:143` always says **"No courses match this filter"**. When `total === 0` there is no filter to blame — the teacher simply has no courses. The filter chips also collapse to just `All` in that state, so the message points at a control that isn't there.

Worse, this is the first thing a new teacher sees, and it offers no way forward: no "create your first course" affordance.

**Fix:** branch on `total === 0` vs "filtered to nothing". The genuinely-empty case should read as a first-run state and point at course creation (`lms_create_course` exists; at minimum a `sendFollowUpMessage` prompt).

## Done when

- No widget ever renders a raw user id or `U-` initials.
- Ungraded exams are worded as ungraded, not as an absent score.
- The zero-courses state has its own copy and a way forward, distinct from a filtered-empty result.
- Fixtures `student_progress_roster:default` (row `u-008`, and Grace Okoye for the exam case) and `course_dashboard:empty` render correctly.

guillermoscript:
## What

Three MCP widget states presented *absent* data as a value a teacher could act on — a raw user id where a name goes, an em dash where an ungraded exam goes, and a filter blamed for an empty course list that has no filter. Fixed, along with three adjacent defects found while fixing them.

Closes #568

## Why

Each symptom is the same mistake: a nullable field rendered through a fallback that produces something plausible-looking instead of saying "we don't have this".

**1. A student with no name rendered as their user id.** `profiles.full_name` is nullable and `fetchProfileNames()` deliberately drops blanks, so `student_name: null` is a normal value on every roster. The roster fell back to `student_id.slice(0, 8)` for the name and `id.slice(0, 2)` for the avatar — `u-008` and `U-`. That identifies nobody and reads as corrupted data. `submission-grader` had the identical fallback.

The honest fix is a placeholder, not a better-looking id. Resolving the real person would mean their email, which lives in `auth.users` and needs `auth.admin.getUserById()` — a service-role call these tools cannot make on the caller's RLS-scoped client. So the copy stays generic and, importantly, *looks* generic: muted italic, so a reviewer can see at a glance which rows are missing a name.

**2. "1 exam" beside "—".** The em dash is this widget's glyph for "no data" everywhere else (progress %, last active), so reusing it for an exam awaiting a grade read as *took an exam, scored nothing*.

**3. The empty course list blamed a filter.** One branch served two situations. With `total === 0` there is no filter to blame, the chips have collapsed to a lone `All`, and a first-run teacher was given no way forward.

### Three adjacent defects, fixed here because leaving them makes the fix incoherent

**4. The ungraded-exam state was unreachable from real data.** In `lms_get_student_progress`, `exam_count` was `ex.n`, and `n` was incremented only `if (s.score != null)`. An ungraded submission therefore vanished: the student read as "0 exams", not "1 exam, ungraded". Fixing only the widget would have shipped copy that never renders. `aggregateExamSubmissions()` now counts every submission and averages only the graded ones.

**5. `exam-submissions` declared `student_name: z.string()`** while its tool sends `names.get(...) ?? null`. A nameless student rendered as an empty cell.

**6. `lms_list_courses` reported a hardcoded `total: 0`** on an empty page, discarding the real `count`. With an offset past the end, a teacher who *does* have courses would get `total: 0` — and, after this change, be shown the "create your first course" screen. Fixing the empty state is what made this latent bug user-visible.

## How to QA

No database, RLS, auth or migration surface is touched, so no `db:reset` is needed — these are widget and MCP-tool changes, verified through the dev-only fixture tools.

```bash
cd mcp-server
npm install
MCP_DEMO_WIDGETS=1 npx mcp-use dev --port 3009
```

> The widget HTML embeds `MCP_SERVER_URL` as the origin it loads its own JS from. If that value does not match the port the dev server actually bound to, the screenshot renders **another checkout's** code and your changes appear to do nothing. Pin the port as above and set `MCP_SERVER_URL=http://localhost:3009` in `mcp-server/.env`.

1. **Roster — the two name/exam cases** (fixture row `u-008` has no name; Grace Okoye has one ungraded exam):
   ```bash
   npx mcp-use client screenshot --mcp http://localhost:3009/mcp \
     --tool lms_demo_student_progress_roster variant=default --theme dark --output roster.png
   ```
   Expect *Unnamed student* in muted italic with a `·` avatar — no `u-008`, no `U-` — and Grace Okoye reading **Ungraded** over "1 exam", not `—`.
2. **Course dashboard — first run:**
   ```bash
   npx mcp-use client screenshot --mcp http://localhost:3009/mcp \
     --tool lms_demo_course_dashboard variant=empty --theme dark --output empty.png
   ```
   Expect "No courses yet", a line of orientation, a **Create my first course** button, no mention of a filter, and no lone `All` chip. Repeat with `--theme light`.
3. **Course dashboard — the filtered-empty case still works:** open `lms_demo_course_dashboard variant=default` in the inspector and click a status chip; every chip matches at least one course, so the grid should never empty out. The "No courses match this filter" branch is what catches a chip that matches nothing, with a **Show all N courses** reset.
4. **Regression on the two other widgets that show a student:** `lms_demo_exam_submissions variant=default` and `lms_demo_submission_grader` — names render exactly as before (every fixture row has one).
5. `cd mcp-server && npm test && npm run build`.

## Screenshots / GIF

Posted as a comment below — before/after for the roster and the course-dashboard empty state, in both themes and in Spanish, plus two A/B GIFs.

## Checklist

- [x] `npm run typecheck` and `npm run test:unit` pass — run as `mcp-server`'s own `npm run build` (which ends in a full type check) and `npm test`: **66 tests pass** (10 of them new here; the rest arrived with #578). The root-package scripts do not cover `mcp-server/`, which is not an npm workspace.
- [x] `npm run build` passes — in `mcp-server/`; 18 widgets built, type check clean.
- [x] Every new tenant-scoped query filters by `tenant_id` — no query was added or changed in shape; `lms_list_courses` keeps its existing `.eq("tenant_id", …)` and only the reported `total` changed.
- [x] Tested with every relevant role — these are teacher/admin-facing widgets, exercised through the dev fixture tools, which is the only way to reach the empty and nameless states deterministically. The roster and dashboard tools are already role-gated in `src/tool-policy.ts`; that gating is untouched.
- [x] Loading and error states handled — the `isPending` branches are unchanged; this PR adds the missing *empty* and *absent-value* states.
- [x] New UI strings added to both `messages/en.json` and `messages/es.json` — **equivalent done, different file.** MCP widgets render outside the Next.js app and have no access to `next-intl`; since #570 / PR #578 they carry their own `STRINGS` tables (`resources/shared/i18n.tsx`). Every string this PR adds — the placeholder name, "Ungraded", and all four empty states — is defined in both `en` and `es` and verified in a Spanish render (below).
- [x] Migration — none.

## Reviewer notes

- **`exam_count` changes meaning** — from "graded submissions" to "submissions". It is read only by this widget and this tool's own summary line; no sort, no at-risk rule, and no other tool consumes it. Worth a look because it changes what a number *means*, not just how it is formatted.
- **The new tests were not executed against the pre-fix tree**, because `aggregateExamSubmissions()` and `student-display.ts` did not exist before this branch. What they pin down is the behaviour the old code provably lacked: the old reducer created its accumulator but only ever incremented on a non-null score, so the ungraded-only case came out as `exam_count: 0` — the assertion of `submitted: 1` could not have passed against it.
- **`submission-grader/widget.tsx` is also touched by #567 / PR #574.** The change here is one line (the raw-id fallback) and is independent of that PR's `?? 0` work, but the two will want a look if they land close together.
- The "no courses on this page" branch exists only because of fix 6 above; without a real `total` it was unreachable, and with the old copy it would have blamed a filter for a pagination overshoot.

### Rebased onto master 2026-07-28

PR #578 (#570, widget UX polish) and PR #576 landed first and touch all four of these widgets, so this branch was rebased rather than merged. Three things changed in the process, all visible in the diff:

- **The new copy is now bilingual.** #578 introduced the per-widget `STRINGS` tables, so `unnamedStudent`, `ungraded`, and the four empty states are defined in `en` and `es` instead of being hardcoded English.
- **`studentDisplayName(name, unnamedLabel?)` takes the placeholder as an argument**, defaulting to English. The placeholder is a widget-owned string and follows the reader's locale; a real `full_name` is database content and is still rendered verbatim.
- **The exam cell composes with #578's severity colours** — it defers to `fmt.percent` + `textClass` for a real average and only overrides them for the ungraded state.

Re-verified after the rebase: `npm test` 66 passed, `npm run build` type check clean, and the roster and dashboard re-shot in `en` and `es` (screenshots in the comment below refreshed).


guillermoscript:
## Visual evidence

Captured with the dev-only fixture tools (`MCP_DEMO_WIDGETS=1`) against the two variants named in the issue's *Done when*, after the rebase onto #578. "Before" was captured on the untouched tree.

### 1 + 2. Roster — nameless student and ungraded exam

Row `u-008` (no `full_name`) and Grace Okoye (one submission, not yet graded).

| Before | After |
|---|---|
| roster before | roster after |

roster A/B

`u-008` → *Unnamed student* in muted italic with a `·` avatar; Grace Okoye's `—` → **Ungraded** over "1 exam". Every named row is untouched — the progress/exam colours come from #578, not from this PR.

Light theme: [before](https://raw.githubusercontent.com/guillermoscript/lms-front/master/docs/qa/568/before-roster-light.png) · [after](https://raw.githubusercontent.com/guillermoscript/lms-front/master/docs/qa/568/after-roster-light.png)

### 3. Course dashboard — the true-empty state

| Before | After |
|---|---|
| dashboard before | dashboard after |

dashboard A/B

"No courses match this filter" (beside a lone, useless `All` chip) → a first-run state with a way forward. The chip row and the "0 courses total" line are hidden, since neither says anything true here.

Light theme: [after](https://raw.githubusercontent.com/guillermoscript/lms-front/master/docs/qa/568/after-dashboard-empty-light.png)

### Spanish

Every string added here is defined in `en` and `es` (`STRINGS` tables, per #578) and rendered via `lang=es`:

| Roster | Dashboard |
|---|---|
| roster es | dashboard es |

*Estudiante sin nombre* · *Sin calificar* · *Todavía no tienes cursos* · *Crear mi primer curso*.

### Regression check

`exam-submissions`, whose schema changed to `student_name: z.string().nullable()` — every fixture row has a name and all render as before:

exam submissions

<sub>Media lives in `docs/qa/568/` on `master`. This comment replaces an earlier one whose links pointed at the now-deleted branch.</sub>

guillermoscript:
Merged as `8119efe` (squash) into `master`; branch deleted.

**Shipped**

- `resources/shared/student-display.ts` — one place decides how an unnamed student renders. `studentDisplayName` / `studentInitials` / `isNamedStudent`, used by `student-progress-roster`, `submission-grader` and `exam-submissions`. No widget derives anything shown to a human from a user id any more.
- The roster's exam column distinguishes "has not sat an exam" (`—`) from "sat one, waiting on a grade" (**Ungraded**, amber).
- `aggregateExamSubmissions()` in `src/tools/analytics.ts` counts every submission and averages only the graded ones, so that second state is reachable from real data at all. **`exam_count` changed meaning**: submissions, not graded submissions.
- `course-dashboard` has four empty states instead of one — no courses at all (first-run copy plus a create-course action, chip row hidden), no courses of a server-filtered status, a chip that matched nothing (with a reset), and a page past the end of the list.
- `exam-submissions` schema corrected to `student_name: z.string().nullable()`, matching what its tool has always sent.
- `lms_list_courses` reports the real `count` on an empty page instead of a hardcoded `0`.

**Verification at merge:** `npm test` 66 passed (10 new), `npm run build` type check clean, CI `verify` green. Roster and dashboard re-shot in light, dark and Spanish after the rebase onto #578 — evidence in the comment above, now served from `docs/qa/568/` on `master`.

**Worth remembering.** Widget screenshots are only as trustworthy as the port: the dev HTML embeds `MCP_SERVER_URL` as the origin it loads its own JS from, and `mcp-use dev` ignores `PORT` and auto-picks a free one. In a worktree that silently renders *another checkout's* code — an "after" shot identical to the "before" one, with nothing in the logs. `npx mcp-use dev --port <n>` with a matching `MCP_SERVER_URL` is the fix.

Issue #604, opened 2026-08-05
Capability-gate the Stripe-only billing operations (customer portal, proration preview)


guillermoscript:
# Capability-gate the Stripe-only billing operations (customer portal, proration preview)

Part of #600. Depends on the schema and routes sub-issues.

## Problem

Four school-facing billing actions call the Stripe SDK directly, so they have no defined behaviour for a school billed through anything else:

| Action | Stripe call |
|--|--|
| `previewPlanChange` | `stripe.subscriptions.retrieve` (`app/actions/admin/billing.ts:445`), `stripe.invoices.createPreview` (482) |
| `changePlan` | `stripe.subscriptions.update` (537) |
| `cancelSubscription` | `stripe.subscriptions.update` (616) |
| `reactivateSubscription` | `stripe.subscriptions.update` (687) |

Plus `app/api/stripe/billing-portal/route.ts:55` (`stripe.billingPortal.sessions.create`), the only "manage my subscription" surface in the product, and `lib/payments/platform-plan-change.ts:148-153`.

Two of these are genuinely provider-specific abilities, and the codebase already has the idiom for that: `ProviderCapabilities` (`lib/payments/types.ts:~155`), whose whole contract is *"lets the app branch on ability, never on provider identity"*. `supportsPlanChange` already distinguishes a native item swap from an app-side supersession — this issue extends the same idea to the two remaining Stripe-only assumptions.

Today a school on Lemon Squeezy or manual transfer would either see a dead "Manage billing" button or hit an unhandled Stripe error, depending on the action.

## Wanted

**Two new capabilities** on `ProviderCapabilities`, with the static `PROVIDER_CAPABILITIES` table and each provider class kept in step (the table's own comment requires this):

- `supportsCustomerPortal` — provider hosts a self-serve subscription-management page. `true` for Stripe; `false` for Lemon Squeezy (dashboard-managed), PayPal, manual, crypto. When `false`, the billing page renders the in-app cancel / change-plan / renewal controls it already has instead of the portal button. The button must not render at all rather than render and fail.
- `supportsProrationPreview` — provider can quote the mid-period delta before committing. `true` for Stripe only. When `false`, `previewPlanChange` returns an honest "change takes effect on <date>, no proration" preview rather than a fabricated number. There is already a note at `lib/payments/platform-plan-change.ts:201-203` about a comped-plan case where reading Stripe's real price mattered — preserve that behaviour on the Stripe path exactly.

[...]

Issue #602, opened 2026-08-05
Nothing in the repo ever writes a platform plan's price id — every card upgrade 400s


guillermoscript:
# Nothing in the repo ever writes a platform plan's price id — every card upgrade 400s

Part of #600.

## Problem

`platform_plans.stripe_price_id_monthly` / `stripe_price_id_yearly` are **read by three call sites and written by none**. Full-repo grep (`app/`, `lib/`, `components/`, `scripts/`, `supabase/`, excluding `lib/database.types.ts`):

**Readers**
- `app/api/stripe/checkout-session/route.ts:79-81` → 400 `"Stripe price not configured for this plan. Please contact support."` at lines 83-88 when unset.
- `app/actions/admin/billing.ts:429,439` (`previewPlanChange` / `changePlan`).
- `lib/payments/platform-plan-change.ts:59,79,148-149` (portal-initiated plan change; resolves the target plan by matching the price id).

**Writers** — none. Re-verified 2026-08-05 across `app/`, `lib/`, `components/`, `scripts/`, `supabase/migrations/`, `supabase/seed.sql`, `supabase/seed-prod.sql`, `supabase/functions/`, `supabase/snippets/`:
- `supabase/migrations/20260217040000_platform_billing.sql:14-15` declares the columns; line 155 seeds the five plans without them, and no later migration touches them.
- `supabase/seed-prod.sql` — the `INSERT INTO platform_plans (...)` column list omits both. **So a freshly provisioned production database has no purchasable plan at all.**
- `app/actions/platform/plans.ts:28` — `updatePlatformPlan`'s `updates` type is `{ name, price_monthly, price_yearly, transaction_fee_percent, limits, features, is_active, sort_order }`. No price-id field; `createPlatformPlan` (line 54) likewise.
- `app/[locale]/platform/plans/plan-editor.tsx` — the super-admin editor renders Name, Sort Order, Monthly Price, Yearly Price, Txn Fee %, Limits/Features JSON (lines 115-163). **No price-id input exists**, so there is no UI path either.
- Nothing in `scripts/`, no edge function, no snippet.

The only place the value is ever meant to be set is a **manual instruction to a human**: `docs/MONETIZATION.md:344` — *"Store price IDs in `platform_plans.stripe_price_id_monthly` / `stripe_price_id_yearly`"*. That step has never been automated, asserted, or seeded.

[...]

Issue #601, opened 2026-08-05
Schema: make platform-billing tables provider-agnostic


guillermoscript:
# Schema: make platform-billing tables provider-agnostic

Part of #600. **Blocks every other sub-issue in the epic.**

> **Pre-launch project.** There is no deployed app and no real users — the same standing note #540 opens with: *"Breaking changes are fine — do not write backfill dances, compatibility shims or phased rollouts for existing rows. Drop and recreate. Prefer the correct end state over the migration-safe one."* This issue is written to that rule. The only rows that exist are seed rows and a handful of test tenants on the cloud project.

## Problem

`supabase/migrations/20260217040000_platform_billing.sql` models school billing around one provider's primitives, so there is nowhere to put a second provider's identifiers:

- `platform_plans.stripe_price_id_monthly` / `stripe_price_id_yearly` (lines 14-15) — two columns, one provider, one interval each. A plan cannot hold both a Stripe price and a Lemon Squeezy variant.
- `platform_subscriptions.stripe_subscription_id` / `stripe_customer_id` (lines 32-33) and `payment_method VARCHAR(50) NOT NULL DEFAULT 'stripe' CHECK (payment_method IN ('stripe','manual_transfer'))` (lines 36-37) — a two-value enum standing in for what the student side already models as an 8-value provider slug (`PaymentProvider`, `lib/payments/types.ts:6`). `idx_platform_subscriptions_stripe` (line 86) indexes the Stripe column only.
- `tenants.stripe_customer_id` (line 75) — a single customer id with the provider implied by the column name.

The student-side tables already solve this the way we want, and #280 renamed them for exactly this reason: `subscriptions` uses `payment_provider` + `provider_subscription_id`, `plans` uses `payment_provider` + `provider_price_id`. Platform billing should read identically to anyone who knows the other half of the codebase.

## Wanted

One migration. Rename and drop outright — no dual-write phase, no compatibility columns, no rollback dance.

**`platform_plan_prices`** (new)
```
plan_id           uuid  → platform_plans(plan_id) on delete cascade
payment_provider  text  -- PaymentProvider slug; same vocabulary as lib/payments/types.ts:6
interval          text  CHECK (interval IN ('monthly','yearly'))
provider_price_id text  NOT NULL
currency          currency_type NOT NULL DEFAULT 'usd'
amount            numeric  -- what the provider actually charges; may differ from
                           -- platform_plans.price_monthly on non-USD rails
[...]

Issue #600, opened 2026-08-05
EPIC: Provider-agnostic platform billing — let a school pay the platform through any provider, not just Stripe


guillermoscript:
# EPIC: Provider-agnostic platform billing — let a school pay the platform through any provider, not just Stripe

## Summary

The **student → school** loop is provider-agnostic: 8 providers behind `IPaymentProvider` (`lib/payments/types.ts:434`), a static `PROVIDER_CAPABILITIES` matrix so the app branches on *ability* never provider identity, one normalized event type (`NormalizedBillingEvent`, `lib/payments/types.ts:378`), one applier (`dispatchBillingEvent`, `lib/payments/webhook-dispatch.ts:60`), and one unified endpoint (`app/api/payments/webhook/[provider]/route.ts`).

The **school → platform** loop (school billing / `platform_subscriptions`) never got that treatment. It is hard-wired to Stripe at every layer: schema, routes, and server actions. Two consequences, both code-verified 2026-08-05:

1. **A school that cannot get a Stripe account cannot pay us at all**, except by offline bank transfer plus a manual super-admin confirmation (`confirmManualPayment`, `app/actions/admin/billing.ts:277`). For LATAM creators — an explicitly targeted segment (`currency_type` carries `mxn/cop/clp/pen/ars/brl`, `docs/MONETIZATION.md`) — that is the common case, not the edge case.
2. **The card path is currently unreachable for every plan anyway** (see sub-issue on plan prices): nothing in the repo ever writes `platform_plans.stripe_price_id_monthly/yearly`, so every card upgrade dies on `"Stripe price not configured for this plan"`. The Stripe-shaped schema is what hid this — a provider-agnostic price table with an admin UI makes the same class of bug impossible to ship silently.

This epic makes the platform-billing loop use the machinery the student loop already has. It is a refactor plus one migration, not a new abstraction: no new interface is introduced, `lib/payments` stays the single contract.

[...]

PR #609, opened 2026-08-06
refactor(billing): unified provider-agnostic checkout + webhook for platform billing (#603)

### LINKED ISSUE #603: Unified checkout + webhook routes for platform billing
# Unified checkout + webhook routes for platform billing

Part of #600. Depends on #601.

> Pre-launch project — no users, no deployed subscribers. Breaking cutovers are preferred over phased ones (#540).

## Problem

Platform billing has its own Stripe-shaped route pair that duplicates — and diverges from — the provider-agnostic pipeline the student loop already runs.

**Checkout.** `app/api/stripe/checkout-session/route.ts` calls `stripe.checkout.sessions.create` directly. Every other checkout in the codebase goes through `getPaymentProvider()` (`lib/payments/index.ts:19`) and the provider's own `createCheckout`, capability-gated by `supportsHostedCheckout`.

**Webhook.** `app/api/stripe/platform-webhook/route.ts` (495 lines) switches on Stripe event names — `checkout.session.completed` (160), `customer.subscription.updated` (252), `customer.subscription.deleted` (340), `invoice.payment_failed` (355), `invoice.paid` (433) — and writes `stripe_subscription_id`, `stripe_customer_id`, `payment_method: 'stripe'` inline (198-201, 218, 367, 445). It does not use `NormalizedBillingEvent` (`lib/payments/types.ts:378`) or `dispatchBillingEvent` (`lib/payments/webhook-dispatch.ts:60`).

The student side already has the pattern to copy, including the reasoning: `app/api/payments/webhook/[provider]/route.ts` documents the pipeline as `raw body → provider.verifyWebhook → persist to webhook_events (idempotent) → provider.normalizeWebhookEvent → dispatchBillingEvent → 200`, with an explicit `SUPPORTED` allowlist so providers without a signed webhook (`manual`, `solana`) are never exposed as an unauthenticated mutation surface.

Cost of the divergence: any lifecycle rule — grace periods, `past_due` handling, the reminder ladder — has to be implemented twice and can silently drift between the two loops. #547 is the precedent for what that costs (two shipped screens differing by the entire platform fee).

## Wanted

**`POST /api/billing/checkout`** — replaces `/api/stripe/checkout-session`.
- Same guards as today (auth, active admin membership via `tenant_users`, existing-subscription block at lines 44-60, free-plan reject at 74-76).
- Resolve the provider: explicit body param, else the tenant's configured platform-billing provider, else the one active `platform_plan_prices` row for the plan/interval. Reject with a specific error when the plan has no price for that provider — not the current catch-all string.
- Require `supportsHostedCheckout`; get or create the customer through `tenant_billing_customers` rather than `tenants.stripe_customer_id`.
- Keep locale-preserving success/cancel URLs (`resolveRequestLocale`, lines 118-122) — that behaviour is load-bearing behind the tenant proxy.

**`POST /api/billing/webhook/[provider]`** — replaces `/api/stripe/platform-webhook`.
- Same five-step pipeline as the student route, reusing `webhook_events` (`supabase/migrations/20260615130000_webhook_events.sql`) for idempotency and its own allowlist.
- Extend `NormalizedBillingEvent` only if platform billing genuinely needs an event the student loop lacks; prefer reuse.
- New `dispatchPlatformBillingEvent()` in `lib/billing/` applying normalized events to `platform_subscriptions` / `tenants` / `revenue_splits`. All existing behaviour must survive the move: revenue-split rewrite on plan change, `renewal_reminder_sent_at` reset, downgrade-to-free on cancel (`lib/billing/downgrade-tenant.ts`), `reconcileAccessCutoff` calls, and the comped-plan protection documented at `lib/payments/platform-plan-change.ts:201-203`.

**Cutover.** Delete the two old routes in the same PR — no forwarders, no deprecation window. There are no real subscribers (see the standing pre-launch note in #540), so the only cost is repointing the endpoint URL in the Stripe dashboard, which is a manual step either way. `STRIPE_PLATFORM_WEBHOOK_SECRET` stays valid: same signing secret, new path. Add the new URL to `docs/DEPLOYMENT.md` next to the existing webhook setup instructions.

**The UI must change too, or the new providers are unreachable.** The upgrade surface is binary today and hardcodes the assumption in three places:

- `components/admin/plan-comparison-table.tsx:25-26` exposes exactly two callbacks — `onSelectPlan` (→ Stripe) and `onManualTransfer` (→ bank form). Replace with a payment-method step listing the providers actually configured for that plan (i.e. rows in `platform_plan_prices`) plus manual transfer.
- `app/[locale]/dashboard/admin/billing/upgrade/upgrade-page-client.tsx:131` — `onManualTransfer={activeStripeSub ? undefined : handleManualTransfer}` removes bank transfer once a Stripe subscription exists. **A school whose card starts failing cannot switch to the method it now needs.** Switching *between* providers must be a supported transition: supersede the old subscription (cancel at the provider, per `supportsPlanChange` / capability), then start the new one — the same supersession model #463 already established student-side.
- The `activeStripeSub` prop (line 28-29) should become provider-neutral.

Also give `platform_payment_requests` a provider column. Its `pending → instructions_sent → payment_received → confirmed` ladder is fully implemented (`sendPaymentInstructions`, `app/actions/platform/plans.ts:129`, shipped in #480 — **do not re-implement it**); it should describe any out-of-band settlement, not only a bank wire.

## Acceptance criteria

- [ ] A Stripe platform subscription can be created, upgraded, downgraded, cancelled and reactivated through the new routes, with `platform_subscriptions` rows identical in shape to before.
- [ ] Replaying the same webhook event twice is a no-op (`webhook_events` idempotency), verified by test.
- [ ] `tests/playwright/` platform-billing and plan-change specs pass unchanged; `subscription-lapse.spec.ts` and `plan-change.spec.ts` in particular.
- [ ] Route-level vitest coverage for the new webhook, following the recipe in the #544 tests.
- [ ] No `getStripe()` import remains in either new route.
- [ ] A school admin can see and choose every provider configured for the plan, and the list is driven by `platform_plan_prices`, not a hardcoded pair.
- [ ] A tenant on Stripe can move to manual transfer (and back) without a super admin touching the database — no double-billing, old subscription superseded.
- [ ] No `=== 'stripe'` comparison introduced in either route or the upgrade UI.

guillermoscript:
## What

Replaces platform billing's Stripe-shaped route pair with the provider-agnostic pipeline the student loop already runs, and opens the upgrade UI to every provider a super admin has actually priced.

- `POST /api/billing/checkout` — was `/api/stripe/checkout-session`
- `POST /api/billing/webhook/[provider]` — was `/api/stripe/platform-webhook`

Both old routes are **deleted in this PR**, no forwarders. Part of #600, depends on #601 and #602 (both merged).

Closes #603

## Why

Platform billing duplicated — and diverged from — machinery that already existed. The webhook re-implemented the whole idempotency dance, switched on Stripe event names, and wrote `payment_provider: 'stripe'` inline; the checkout route called `stripe.checkout.sessions.create` directly and never touched `getPaymentProvider()`. Any lifecycle rule — grace periods, `past_due`, the reminder ladder — had to be written twice and could drift silently. #547 is the precedent for what that costs.

The user-facing half mattered as much: the upgrade surface exposed exactly two callbacks (Stripe, bank wire), so a super admin could configure Lemon Squeezy or PayPal and **no school could ever pick it** — and `onManualTransfer={activeStripeSub ? undefined : …}` removed bank transfer once a card subscription existed, i.e. exactly when a failing card made it necessary.

### Two blockers the issue didn't anticipate

Both found while reading the code; flagging them because they changed the design:

**1. `supportsHostedCheckout` is `false` for Stripe.** The issue asks the new checkout route to gate on it — that would have rejected the one provider platform billing works with today. The flag isn't wrong: it describes the *student* checkout shape, where we deliberately use Connect + PaymentIntents/Elements. Stripe Checkout Sessions on the platform account are a different product. It's also read nowhere outside `types.ts` and the provider classes, so it couldn't carry the gate.

Resolved with a precisely-named capability, `supportsPlatformBillingCheckout`, plus `CreateCheckoutParams.hosted`. No new interface — the existing contract gains one flag and one param, which is the epic's stated preference.

**2. The shared Stripe normalizer was lossy for platform billing** — no `checkout.session.completed` case at all, and it dropped the interval, customer id, both period bounds and the cancel schedule. `NormalizedBillingEvent` gains those as **optional, additive** fields plus a normalized `subscriptionStatus`; the student dispatcher ignores every one of them, so its behaviour is unchanged.

## What moved, and what deliberately didn't

`dispatchPlatformBillingEvent()` (`lib/billing/platform-webhook-dispatch.ts`) is the platform twin of `dispatchBillingEvent`. Two dispatchers rather than one because the loops write disjoint tables and disagree about what a row means — student subscriptions are per-user, platform rows are one per tenant with a plan, a split and a tenant mirror. Folding them together would have meant a tenant/user branch inside every case. Everything *above* that layer is shared: the event vocabulary, the five-step pipeline, and the `webhook_events` ledger.

Every prior behaviour survives the move: the subscription upsert shape, the `tenant_billing_customers` write, the revenue-split rewrite, the out-of-order period guard, the status whitelist, `downgradeTenantToFree` on cancel, the dunning email, the comped-plan protection, and the `renewal_reminder_sent_at` reset.

`applyPortalPlanChange` is now provider-agnostic — a normalized input plus a `revertToPrice` callback built from the provider's own `updateSubscription` (gated on `supportsPlanChange`). That's what keeps `getStripe()` out of both new routes entirely.

**Idempotency is namespaced `platform:<provider>`.** The student Connect route logs under plain `stripe`; both endpoints can be registered for the same event types on the *same* Stripe account, so one shared key space would let whichever route ran first mark an event processed and make the other skip work it had not done.

## Bugs found and fixed along the way

- **Stale checkout metadata.** Providers echo creation-time metadata on *every* later event for the life of a subscription. Trusting it past the first activation would rewrite a stale `plan_id` over a legitimate portal plan change **and skip the reconciliation entirely**. The plan now comes from metadata only while the tenant has none on file; from then on the price is the source of truth. Caught by a test, pinned by one.
- **Triple dunning email.** One failed charge produces several `past_due` events (the invoice *and* the subscription status change), plus any redelivery. The email now fires only on the transition into dunning.
- **Dropped pre-clover period fallback.** My first pass read the period only off `items.data[0]`. Any account pinned to an older API version — which is exactly what `stripe listen` replays — would activate a plan with **no period at all**, and the expiry cron reads that column. Found by running it, not by reading it.
- **Dialog overflow.** `Button` carries `whitespace-nowrap`, so the bank-transfer description set a one-line intrinsic width and pushed every option past the panel edge. Found in the first screenshot.

## How to QA

On a fresh `npm run db:reset`, as `owner@e2etest.com` / `password123` on `default.lvh.me:3000`.

**The payment-method step (no Stripe credentials needed):**

1. Go to **Settings → Billing → Upgrade** (`/en/dashboard/admin/billing/upgrade`).
2. Click **Choose plan** on Pro. A dialog asks *"How would you like to pay?"* and lists **Stripe** (from the seeded `platform_plan_prices` row) plus **Bank transfer**.
3. As a super admin, add a Lemon Squeezy price for Pro under **Platform → Plans**, reload, and open the dialog again — Lemon Squeezy now appears. Nothing was hardcoded; deactivate the row and it disappears.
4. Confirm no student-only rail (Solana, Binance) is ever offered — those are filtered on `supportsPlatformBillingCheckout`, not on a slug list.
5. **Bank transfer** is present whatever the school's current method is, including for a school already on a card. On the *current* plan card, a **Change payment method** button now exists — previously there was no route to a different rail without a plan change.

**Full loop against real Stripe test mode** (what I ran — needs a real test price id):

```bash
stripe listen --forward-to lvh.me:3000/api/billing/webhook/stripe \
  --events checkout.session.completed,customer.subscription.created,customer.subscription.updated,customer.subscription.deleted,invoice.paid,invoice.payment_failed
# put the printed whsec_… in STRIPE_PLATFORM_WEBHOOK_SECRET, restart dev
# replace the seeded placeholder price id with a real test-mode one
```

Then click through the dialog to Stripe and complete the subscription.

**Automated:** `npx playwright test tests/playwright/platform-billing-provider-choice.spec.ts` (5 tests).

## Verified end-to-end against live Stripe test credentials

Not just unit tests — the real flow, with `stripe listen` forwarding to the new path:

| Step | Result |
|---|---|
| `POST /api/billing/checkout` from the UI | `200` + a live `checkout.stripe.com` session; a real Stripe customer was created and persisted to `tenant_billing_customers` |
| Subscription created at Stripe | `platform_subscriptions` → `status=active`, `payment_provider=stripe`, `interval=monthly`, correct `provider_subscription_id` and `current_period_end`; `tenants` → `plan=pro`, `billing_status=active`; `revenue_splits` → 2% / 98% |
| Signed replay of an already-processed event, twice | `200 {"received":true,"duplicate":true}` both times, **zero writes** |
| Deliberately bad signature | `400 Invalid signature` |
| Cancel at Stripe | tenant downgraded to `free`, `billing_period_end` cleared, split back to 10% / 90% |
| `webhook_events` | logged under `platform:stripe`, both events `processed_at` set, no errors |

Error paths, all previously one catch-all string ("Stripe price not configured for this plan"), which is precisely why #602 survived months of passes:

| Request | Response |
|---|---|
| free plan | `Cannot subscribe to free plan via checkout` |
| `interval: weekly` | `Invalid interval` |
| `provider: solana` | `solana cannot be used to pay for a platform plan.` |
| provider with no price row | `This plan has no active monthly price for lemonsqueezy. Configured providers: stripe.` |
| several providers, none chosen | `Several payment providers are configured for this plan (lemonsqueezy, stripe). Choose one.` |

## Screenshots

**Payment-method step — the list is `platform_plan_prices`, not a hardcoded pair.** Lemon Squeezy appears here only because a price row was added for it.

Payment method dialog listing Stripe, Lemon Squeezy and Bank transfer

**The upgrade page itself is unchanged** — the plan cards still lead the screen; only what happens after "Choose plan" is different:

Upgrade page plan cards

The redirect from that dialog lands on a real `checkout.stripe.com` page for "QA 603 Pro Plan — $29,00/month", prefilled with `owner@e2etest.com` (the customer the route created). Not screenshotted here because reproducing it needs a live test-mode price id rather than the seeded placeholder; the table above records what it returned.

## Deploy step — required, and nothing works until it happens

The Stripe dashboard endpoint URL must be repointed from `/api/stripe/platform-webhook` to `/api/billing/webhook/stripe`. **Edit the existing endpoint rather than creating a new one** — that keeps the signing secret, so `STRIPE_PLATFORM_WEBHOOK_SECRET` is unchanged. Until it's done, Stripe posts to a path that no longer exists: a school's payment succeeds and only our side of it goes missing. Documented in `docs/DEPLOYMENT.md`; the symptom is a 404 in Stripe's delivery log.

## Notes for the reviewer

- **`applyPortalPlanChange`'s signature changed**, so `platform-plan-change.test.ts` was rewritten against the normalized input. The assertions about *outcomes* (reverted / applied_over_limit / noop / override-ignored) are unchanged — that's what pins the behaviour.
- **`tests/unit/platform-webhook.test.ts` → `platform-billing-webhook-route.test.ts`** (30 tests). It drives the new route with the **real** Stripe normalizer and the **real** dispatcher against a recording Supabase fake — only the signature check, mailer and two helpers with their own suites are stubbed.
- **One deliberate behaviour change:** an unrecognised provider status (`paused`, `incomplete`, `unpaid`) used to normalize to `null` — silence — leaving a subscription that is demonstrably not paying reading as healthy. It now maps to `past_due`, which is what the Stripe-shaped platform route always did and which revokes no access on either side.
- The billing portal (`/api/stripe/billing-portal`) is still Stripe-only — that's #604, deliberately out of scope.
- `docs/PLATFORM_ADMIN.md` claimed the platform webhook calls `rewardReferrer()` on `invoice.paid`. It never did — grep finds one definition and no call site. Corrected rather than carried forward.

## Checklist

- [x] `npm run typecheck` and `npm run test:unit` pass — **716 unit tests across 56 files**
- [x] `npm run build` passes
- [x] Tested with every relevant role — admin (the only role that can reach billing); non-admin members get 403, anonymous 401, both covered by tests
- [x] Loading and error states handled — the dialog disables during checkout, every failure path returns a specific message
- [ ] New UI strings added to both `messages/en.json` and `messages/es.json` — **not done**: the new dialog follows `plan-comparison-table.tsx`, which is hardcoded English throughout. Translating one dialog inside an untranslated screen would look worse than leaving it consistent; the whole surface deserves one pass, and I did not want to hide that in a billing refactor.
- [x] Migration applies cleanly on `npm run db:reset`, and `lib/database.types.ts` was regenerated — `platform_payment_requests.payment_provider`, CHECKed against the same eight slugs as the #601 tables. Applied and exercised locally. **RLS unchanged** — it adds a column to a table whose policies already exist and are unaffected. **Not yet applied to cloud.**

https://claude.ai/code/session_01DGGk17ggmBAdJG7Eca981U


guillermoscript:
Merged as `597edfff` and **deployed to production**, with the migration and the Stripe endpoint both handled.

**Deploy steps, all done:**

| Step | State |
|---|---|
| Cloud migration | Applied to production Supabase. `platform_payment_requests.payment_provider` — `NOT NULL DEFAULT 'manual'`, the 8-slug CHECK, and the `(payment_provider, status)` index. The one existing row defaulted to `manual`. |
| Migration ledger | Corrected `20260806013716` → `20260806120000`. The MCP tool stamps its own timestamp, and that drift is what hides the *next* missing migration — so the cloud ledger now matches the repo filename. |
| App deploy | `deploy.yml` run [31063915592](https://github.com/guillermoscript/lms-front/actions/runs/31063915592) green → ghcr → Dokploy. Nothing under `mcp-server/**` changed, so `deploy-mcp.yml` was correctly not involved. |
| Stripe endpoint | `we_1Tak9wF3QAnraq1I3GVDkEeX` repointed to `/api/billing/webhook/stripe`. **Edited in place**, so the signing secret is unchanged and `STRIPE_PLATFORM_WEBHOOK_SECRET` needed no update. Same six events. |

**Verified against production after the deploy** — the two paths swapped exactly as intended:

```
POST /api/stripe/platform-webhook  → 404              (was 400; route is gone, no forwarder)
POST /api/billing/webhook/stripe   → 400 {"error":"Invalid signature"}
POST /api/billing/webhook/solana   → 404 {"error":"Unknown provider: solana"}
POST /api/billing/checkout         → 401              (unauthenticated)
```

The `solana` 404 is the allowlist doing its job: a provider with no signed webhook never gets an endpoint, because that endpoint would be an unauthenticated way to activate a school's plan.

**Two things worth knowing, neither introduced by this PR:**

1. That Stripe endpoint is **`disabled`** and `livemode: false`. Platform billing is not receiving live Stripe events today. I repointed the URL but deliberately did not enable it — that is a launch decision, not a deploy step.
2. If production ever runs a **live-mode** key, there is a separate live endpoint that still points at the old path. It needs the same in-place URL edit, and a test-mode key cannot see or change it.

Issue #546, opened 2026-07-25
School billing lifecycle: irreversible cancel, stale renewal requests pause expiry forever


guillermoscript:
Part of EPIC #540. §2.3.

Five independent defects in the school→platform billing lifecycle. All verified by reading the code; none has any test coverage (`grep -rl "platform-webhook\|expire-platform-subscriptions" tests/` returns nothing).

## 1. Cancel is irreversible, and paying again does not undo it 🔴

`app/actions/admin/billing.ts:580-590` — `cancelSubscription()` sets `cancel_at_period_end: true` + `canceled_at` on the manual-transfer branch.

`app/actions/admin/billing.ts:324-338` — `confirmManualPayment()`'s upsert writes `status`, `interval`, both period columns, `grace_period_end` and `renewal_reminder_sent_at`, but **not** `cancel_at_period_end` or `canceled_at`. PostgREST `ON CONFLICT DO UPDATE` only touches supplied columns, so the stale `true` survives.

Then `app/api/cron/expire-platform-subscriptions/route.ts:176-183` (Phase 4) selects `status='active' AND cancel_at_period_end=true AND current_period_end < now` and downgrades to free — while Phases 1 and 2 both filter `.eq('cancel_at_period_end', false)`, so that row gets **no renewal reminder and no grace window**.

Net: a school cancels, changes its mind, pays for a full year, has it confirmed, and is **silently dropped to free at the end of that year**, having received no warning at all.

Same shape on Stripe: `changePlan()` (`:503-512`) sends only `items`/`proration_behavior`/`metadata` and never sets `cancel_at_period_end: false`, while `resolvePlatformPlanChange` (`:386`) accepts a still-`active` sub. `checkout-session/route.ts:52-56` blocks re-checkout while that sub is active, and `billing-dashboard-client.tsx` imports only `uploadPaymentProof`, `requestManualRenewal` and `cancelSubscription` — **there is no reactivate action in the UI at all.**

## 2. A never-paid renewal request pauses the expiry downgrade forever 🔴

- `expire-platform-subscriptions/route.ts:36` — `PENDING_RENEWAL_STATUSES = ['pending','instructions_sent','payment_received']`; `:152-163` — any matching row makes Phase 3 `continue`, with **no timeout, no re-notification, no re-arming of `grace_period_end`**.
[...]

Issue #545, opened 2026-07-25
Student subscription state machine: cancel_at is unusable, plan switches fail or are free


guillermoscript:
Part of EPIC #540. §2.2.

### 1. `cancel_at` is `NOT NULL DEFAULT now()`, so the Solana crank cancels every subscription instead of renewing it 🔴

Live DDL (verified, also `20260126190500_lms_complete.sql:1907`):

```
cancel_at | NOT NULL | DEFAULT timezone('utc'::text, now())
```

No writer sets it at creation. `handle_new_subscription` (`20260719120000_guard_parallel_subscriptions.sql:75-82`) omits it; `change_subscription_plan`'s INSERT (`20260719170000:135-144`) omits it. **Every row is born with `cancel_at` permanently in the past.**

`lib/payments/solana-pull-decision.ts:82-87`:

```ts
const cancelDue = row.cancel_at ? BigInt(...) <= now : false
if (row.cancel_at_period_end || cancelDue) return { action: 'cancel', ... }
```

So the first crank run after a period rolls takes the **cancel** branch: status → `canceled`, `handle_subscription_status_change` expires the entitlements, no pull is submitted. **Native `solana_subs` recurring billing can never renew.** A paying student loses access at period end, the school is never paid for period 2 onward, and the on-chain delegation stays live.

Why it shipped green: `tests/unit/solana-pull-decision.test.ts:29` fixes `cancel_at: null` — a value the NOT NULL column cannot hold.

Both reactivate paths (`app/actions/subscriptions.ts:158-168`, `app/actions/admin/subscriptions.ts:167-174`) already document `cancel_at` as a non-nullable trap; neither fixes the creation-time default.

### 2. `change_subscription_plan` writes `cancel_at = NULL` into that NOT NULL column 🔴

`supabase/migrations/20260719170000_change_subscription_plan.sql:124`, on the "reactivate an existing `(user_id, new_plan_id)` row" branch — which is taken whenever such a row exists in **any** status, i.e. exactly the state step 1 of the same function leaves behind (old row set to `canceled`, never deleted).

So **every switch back to a previously-held plan fails with `23502`**. The canonical Basic → Pro → Basic downgrade always aborts. `mapPlanChangeError` (`app/[locale]/(public)/checkout/actions.ts:247-253`) matches no known token, so the student sees the generic "Could not change your plan. Please try again." forever — after the Stripe price swap has already been applied and then reverted by the compensation block, leaving them on the old plan with no explanation.

[...]

Issue #461, opened 2026-07-18
Fix platform downgrade limit-block leaving Stripe/DB plan desynced


guillermoscript:
Part of #458 (Phase 1 — billing correctness).

## Problem
`app/api/stripe/platform-webhook/route.ts` (`customer.subscription.updated`, ~lines 150-168): when a school downgrades via the Stripe Customer Portal while over the target plan's course/student limits, the webhook logs an error and skips the DB update — but **leaves the Stripe subscription on the new cheaper price** (code comment: "Don't update the plan — leave Stripe subscription active"). Result: school pays for starter, DB keeps pro, nobody is notified.

## Wanted
Pick one consistent behavior:
1. **Revert**: update the Stripe subscription back to the old price and notify the admin why the downgrade was rejected; or
2. **Apply**: accept the downgrade, mark the tenant over-limit (read-only/grace state on excess courses/students), and notify.

Either way: send an email/notification to school admins and record the event. Long-term this is pre-empted by the in-app downgrade flow (#458 Phase 3) that checks limits *before* touching Stripe.

## Acceptance
- [ ] Stripe price and `tenants.plan`/`platform_subscriptions.plan_id` can never disagree after this webhook.
- [ ] Admin gets notified on a blocked/adjusted downgrade.
- [ ] Unit test for the over-limit path.

guillermoscript:
🤖 An AI agent (Claude Code, driven by @guillermoscript) is picking this issue up. Plan of attack below — comments welcome.

## Root cause / current behavior

`app/api/stripe/platform-webhook/route.ts` (`customer.subscription.updated`, lines 141–189) handles plan changes made through the Stripe Customer Portal:

1. It looks up the new plan with `.select('slug, plan_id, transaction_fee_percent')` — **`limits` is never selected**, so the downgrade guard that reads `newPlan.limits` always sees `undefined`, defaults to `-1` (unlimited), and **never fires**. Today every portal downgrade is applied to the DB regardless of course/student limits — the guard is dead code.
2. If the guard *did* fire (the bug this issue describes), it `break`s without touching the DB while Stripe stays on the new cheaper price — school pays for `starter`, DB keeps `pro`, nobody is told.

Either way the invariant "Stripe price ⇔ `tenants.plan`/`platform_subscriptions.plan_id` agree" is not enforced.

## Approach — Revert behavior (option 1 from the issue)

[...]

PR #612, opened 2026-08-06
refactor(billing): capability-gate the Stripe-only billing operations (#604)

### LINKED ISSUE #604: Capability-gate the Stripe-only billing operations (customer portal, proration preview)

guillermoscript:
## What

Adds two capabilities to `ProviderCapabilities` — `supportsCustomerPortal` and `supportsProrationPreview` — moves the last Stripe SDK calls in the school-billing surface behind `IPaymentProvider`, and repoints `POST /api/stripe/billing-portal` to `POST /api/billing/portal`.

Closes #604

## Why

Four school-facing billing actions plus the "manage billing" route called the Stripe SDK directly and branched on `payment_provider === 'stripe'`, so a school billed through anything else had no defined behaviour.

The live victim is **Lemon Squeezy**, which `supportsPlanChange` already lets through the plan-change dialog. Reproduced on `master` before touching anything (before/after screenshots in the comment below):

- The dialog promised *"Review the proration before confirming. **Your card on file is charged automatically.**"* — a Stripe sentence, to a school that has no card on file with us.
- It then showed *"We couldn't load a price preview, but you can still switch — your next invoice will reflect the change."* That reads as a transient hiccup. It was not: `previewPlanChange` had thrown on the hardcoded Stripe guard, and Lemon Squeezy has no quote API at all.
- Confirming produced the toast **"In-app plan change is only available for Stripe subscriptions."** A dead end, reached only after two screens of encouragement.

`cancelSubscription` was quieter and worse: the `payment_provider === 'stripe'` branch meant a Lemon Squeezy subscription was cancelled **only in our database** while Lemon Squeezy kept billing the school.

Two notes on the issue body, both of which reduced the work:

- **`lib/payments/platform-plan-change.ts` has no Stripe references.** #603 already made it provider-agnostic (`applyPortalPlanChange` takes an injected `revertToPrice`), so the line numbers in the issue predate that refactor. The comped-plan override guard the issue asked us to preserve is provider-neutral already and is untouched.
- **`updateSubscription` already existed** on `IPaymentProvider` and `StripePaymentProvider` with exactly the item-swap semantics `changePlan` was open-coding, so that action collapsed to a call rather than a new method.

## Changes made

**Two new capabilities**, set explicitly in the static `PROVIDER_CAPABILITIES` table and in all 8 provider classes:

- `supportsCustomerPortal` — the provider hosts a self-serve management page we can mint a session URL for. `true` for Stripe only. False for Lemon Squeezy as well: it has a customer portal, but it is reached from LS's own dashboard and emails rather than a URL we mint, so there is nothing for `createCustomerPortalSession` to return.
- `supportsProrationPreview` — the provider can quote the mid-period delta before we commit. `true` for Stripe only. Deliberately separate from `supportsPlanChange`: Lemon Squeezy performs the swap and prorates on its own side, but exposes no "what would this cost" call, so we can change the plan and still be unable to quote it.

**Three outcomes where there used to be two.** `previewPlanChange` now distinguishes a real quote, a rail that *cannot* quote (`noProration`, carrying the date the plan renews), and a quoting provider whose call failed (unchanged best-effort behaviour). The dialog renders the middle case as "This change takes effect immediately and your plan renews on 8/31/2026. Your payment provider doesn't quote a mid-period amount, so no proration is shown here", and drops the "your card on file is charged automatically" promise for it.

**SDK calls moved into `StripePaymentProvider`**: `previewSubscriptionChange` and `createCustomerPortalSession`, plus `cancelAtPeriodEnd` on the existing `updateSubscription`. `app/actions/admin/billing.ts` resolves its provider from `platform_subscriptions.payment_provider` and holds no `Stripe` object anywhere.

**Portal route** moves to `app/api/billing/portal/route.ts` (the `/api/billing/*` namespace #603 established), resolves the provider from the subscription, returns 400 `portal_unsupported` when the capability is false, and reads `tenant_billing_customers` for *that* provider. The old Stripe-namespaced route is deleted; its single caller and three docs references are updated.

**UI gate**: `getSubscriptionStatus` replaces `hasStripeCustomer` with `canUseCustomerPortal` (capability **and** a customer id for that provider). On a non-portal provider the button does not render at all rather than rendering and failing; cancel / reactivate / change-plan stay reachable.

**Invariants preserved deliberately:**

- The provider is updated **before** any DB write (#461), so its price and our `plan_id` can never disagree.
- A plan change still clears a pending cancellation (#546 §1) — that flag moved into `UpdateSubscriptionParams.cancelAtPeriodEnd` rather than being dropped; without it a school changes plan, pays, and is still downgraded at period end.
- `cancel_at_period_end` remains the only signal that a cancel is scheduled (#545); `canceled_at` stays informational and is written and cleared alongside it. Cancelling still cannot improve a status.
- A `selfManagedPeriod` rail (bank transfer, one-time crypto) makes **no** provider call on cancel or reactivate — there is no subscription object out there to talk to.
- The Stripe proration arithmetic is byte-for-byte the same, just relocated; conversion to major units now happens inside the provider, matching the `NormalizedBillingEvent.amount` convention.

**Out of scope on purpose:** building an app-side plan-change *supersession* for platform billing on providers with `supportsPlanChange: false`. Those schools are still sent to checkout — now with a provider-accurate message instead of one naming Stripe. The student-facing Stripe Connect surfaces (`create-payment-intent`, `connect`, `webhook`, `lib/stripe-connect.ts`) are untouched, which is why the acceptance grep excludes them.

## Type of change

Refactor (with a user-visible correctness fix to the plan-change and cancel flows on non-Stripe providers).

## How to QA

On a fresh `npm run db:reset`, dev server on `lvh.me`. Sign in as `owner@e2etest.com` / `password123` at `default.lvh.me:3000`.

Default School is on the free plan with no subscription, so give it one to test against:

```sql
-- A Lemon Squeezy school: can swap plans in place, cannot quote, no portal.
INSERT INTO platform_plan_prices (plan_id, payment_provider, interval, provider_price_id, is_active)
SELECT plan_id, 'lemonsqueezy', i.interval, 'ls_variant_' || slug || '_' || i.interval, true
FROM platform_plans, (VALUES ('monthly'),('yearly')) AS i(interval)
WHERE slug <> 'free';

INSERT INTO platform_subscriptions (tenant_id, plan_id, payment_provider, status, interval,
  current_period_start, current_period_end, cancel_at_period_end)
VALUES ('00000000-0000-0000-0000-000000000001',
        (SELECT plan_id FROM platform_plans WHERE slug='pro'),
        'lemonsqueezy','active','monthly', now() - interval '5 days', now() + interval '25 days', false);

UPDATE tenants SET plan='pro', billing_status='active' WHERE id='00000000-0000-0000-0000-000000000001';
```

1. Go to **Settings → Billing**. There is **no** "Manage" button and no "View Invoices" link — Lemon Squeezy has no portal we can open. "Change Plan" and "Cancel plan" are both present.
2. In the browser console: `await (await fetch('/api/billing/portal',{method:'POST',headers:{'Content-Type':'application/json'},body:'{"locale":"en"}'})).json()` → `{ code: 'portal_unsupported', ... }` with status 400. Nothing throws.
3. Click **Change Plan** → pick **Business** → choose **Lemon Squeezy (Current method)**. The dialog says the change takes effect immediately, gives the renewal date, and states that the provider doesn't quote a mid-period amount. **No invented numbers, and no "we couldn't load a price preview".** The intro no longer promises a card on file is charged.
4. Click **Confirm change**. Without real Lemon Squeezy credentials this ends at `LemonSqueezy updateSubscription: invalid variant id "ls_variant_business_monthly"` — the point is that the request now reaches the **Lemon Squeezy provider** and is rejected only by the fake variant id, where on `master` it stopped at "In-app plan change is only available for Stripe subscriptions."
5. Click **Cancel plan** and confirm. It succeeds and the overview shows the pending cancellation. (On `master` this silently skipped Lemon Squeezy entirely; with real credentials the provider is now told first.)
6. Flip the same school to Stripe to confirm the gate works in both directions:
   ```sql
   UPDATE platform_subscriptions SET payment_provider='stripe',
     provider_subscription_id='sub_fake_local', provider_customer_id='cus_fake_local'
   WHERE tenant_id='00000000-0000-0000-0000-000000000001';
   INSERT INTO tenant_billing_customers (tenant_id, payment_provider, provider_customer_id)
   VALUES ('00000000-0000-0000-0000-000000000001','stripe','cus_fake_local');
   ```
   Reload Billing — the **"Manage"** button is back.
7. Set the school to bank transfer (`payment_provider='manual'`) and cancel: it still schedules the cancellation with no provider call at all.

Reset with `npm run db:reset` when done.

## Testing

- [x] `npm run typecheck` — clean.
- [x] `npm run test:unit` — **59 files, 787 tests, all passing.**
- [x] `npm run build` — compiled successfully; `/api/billing/portal` appears in the route manifest and `/api/stripe/billing-portal` is gone.
- [x] `npx eslint` on every changed file — **0 errors.** 4 warnings in `app/actions/admin/billing.ts` (unused imports) are pre-existing; verified identical on `master` via `git stash`.
- [x] Acceptance grep from the issue: `grep -rn "getStripe()" app/actions app/api/billing lib/billing` returns nothing.
- [x] **New:** `tests/unit/provider-capability-parity.test.ts` (20 tests) — every provider class's `capabilities` deep-equals its static table entry, every capability key is declared explicitly as a boolean on all 8 providers (a `toEqual` alone would silently pass an undefined key, which is exactly how a new capability goes unnoticed), a provider claiming a capability implements the matching method, and quoting is never claimed without the ability to make the change. This is the drift the table's own comment has warned about since #601.
- [x] **New:** `tests/unit/platform-billing-capability-gates.test.ts` (15 tests) — the preview/cancel/reactivate branch points, driven by the real `PROVIDER_CAPABILITIES`. `@/lib/stripe` is mocked to **throw**, so any action still reaching for the SDK fails the suite.
- [x] `tests/unit/plan-change-action.test.ts`, `plan-change-provider.test.ts`, `platform-billing-checkout.test.ts`, `platform-billing-crypto-rails.test.ts`, `subscription-state-machine.test.ts`, `platform-plan-change.test.ts`, `platform-billing-webhook-route.test.ts` — all pass **unchanged**.
- [x] `StripePaymentProvider.createCustomerPortalSession` exercised against **real test-mode Stripe** (`sk_test_` from `.env.local`): created a customer, minted a live `billing.stripe.com/p/session/test_…` URL, deleted the customer. The account is left as it was found.
- [ ] Playwright specs (`plan-change`, `parallel-subscription-guard`, `subscription-plan-change-rpc`) — **not run in this session.** They cover the student-facing plan change, which this PR does not touch (its `changePlan` lives in `app/[locale]/(public)/checkout/actions.ts` and is pinned by the two unit suites above, both green). Worth a run in CI before merge.

**One test file changed, in substance unchanged:** `platform-billing-lifecycle.test.ts` asserted against the Stripe SDK seam (`stripe.subscriptions.update` calls). Since the actions now drive `IPaymentProvider`, those assertions moved up one level to the provider seam — same invariants (a self-managed rail makes no provider call; a plan change clears a pending cancellation), same count, all 13 passing. Its `@/lib/stripe` mock now throws, for the same reason as above.

## Checklist

- [x] `npm run build` passes
- [x] Tested with every relevant role (admin — this surface is admin-only; `verifyAdminAccess` and the route's membership check are unchanged)
- [x] Loading and error states handled (the portal button now surfaces a toast instead of swallowing a failed response)
- [x] New UI strings added to both `messages/en.json` and `messages/es.json` (`previewIntroNoProration`, `noProration`, `noProrationWithDate`, `portalUnavailable`)
- [x] Migration — none; this PR is code-only.

## Screenshots / GIF

Before/after screenshots and a verification GIF follow in a comment below (private repo, so they are attached rather than linked).


guillermoscript:
## Visual verification

All captures are the same school (`default.lvh.me`, `owner@e2etest.com`) with its platform subscription set to **Lemon Squeezy** — a provider that *can* swap plans in place but cannot quote the change and has no portal we can open. The "before" shots were taken on `master` before any code changed.

### Before — the plan-change dialog on Lemon Squeezy

Two screens of Stripe-shaped encouragement, ending in a refusal.

| The dialog | Confirming it |
|---|---|
| before dialog | before error |

The dialog promises *"Your card on file is charged automatically"* to a school that has no card on file with us, then reports *"We couldn't load a price preview"* — which reads as a transient hiccup but was actually the hardcoded Stripe guard throwing. Confirming gives **"In-app plan change is only available for Stripe subscriptions."**

### After — the same dialog

after dialog

*"This change takes effect immediately and your plan renews on 8/31/2026. Your payment provider doesn't quote a mid-period amount, so no proration is shown here."* No invented amount, no implied failure, and the card-on-file sentence is gone.

Confirming now reaches the Lemon Squeezy provider and is rejected only by the fake variant id this local fixture uses — which is the point: the request routes by capability instead of stopping at a provider-identity check.

after reaches provider

### The portal button, gated both ways

| Lemon Squeezy — no portal button | Stripe — "Manage" is back |
|---|---|
| no portal | stripe portal |

On the left there is no dead button, and **Change Plan** and **Cancel plan** are still reachable — that is the acceptance criterion. Flipping only `platform_subscriptions.payment_provider` to `stripe` brings **Manage** back, so the gate is the capability rather than a leftover customer row.

### Flow recording

verification

Billing page (no portal button, in-app controls present) → upgrade page → payment-method sheet → the honest no-proration dialog.

**Please visually confirm before merging**, in particular that the new copy reads correctly in both English and Spanish (`messages/es.json` carries `previewIntroNoProration`, `noProration`, `noProrationWithDate` and `portalUnavailable`). One thing worth a reviewer's eye that these captures cannot show: the Stripe proration quote itself. Its arithmetic is unchanged and only moved into `StripePaymentProvider`, and `createCustomerPortalSession` was exercised against real test-mode Stripe (a live `billing.stripe.com/p/session/test_…` URL was minted and the test customer deleted), but no local school has a real Stripe platform subscription to render a quote from.

guillermoscript:
Follow-up review of my own diff turned up one defect, now fixed in `f6e5ec15`.

The cancel and reactivate paths gate the provider call on `selfManagedPeriod`, which is right — but the call itself sat inside `if (provider.cancelSubscription)`. A provider that drives its own billing period yet ships no implementation was silently skipped, and the local mirror (`cancel_at_period_end`, `canceled_at`) was written regardless. That is the same DB-only cancellation this issue set out to remove: the school reads "cancelling at period end" while the provider keeps charging. The reactivate direction is worse — the school is told its subscription is safe while the provider is still set to stop billing, so the plan lapses anyway.

`changePlan` and `/api/billing/portal` already throw when a declared capability has no implementation behind it; these two now match.

PayPal is the provider this describes today: `selfManagedPeriod: false`, ships `cancelSubscription`, ships no `reactivateSubscription`. It is unreachable as a platform-billing rail only because `supportsPlatformBillingCheckout` is still `false`, carrying `TODO(#479): flip once proven against real credentials` — so this was latent rather than live, and the throw is what will surface it when #479 flips that flag.

Two tests added to `platform-billing-capability-gates.test.ts` (48 → 50). Both were confirmed to fail against the previous commit and pass against this one, so they pin the behaviour rather than describe it.

Verification after the fix: 789 unit tests across 59 files pass, `tsc --noEmit` clean, eslint 0 errors (4 warnings pre-existing on `master`).

guillermoscript:
## Live Lemon Squeezy verification — the gap in the Testing section is now closed

The PR body notes that the Lemon Squeezy path had only been exercised as far as "our code reached LS and LS answered", because the local `LEMONSQUEEZY_API_KEY` was a placeholder. A real test-mode key and a real subscription are now in place, so all three operations have been run against the live Lemon Squeezy API.

Subscription `2412708` (test mode), starting on variant `1990865`. After each call the subscription was re-read **from Lemon Squeezy** rather than trusting the value our provider returned:

| Step | State Lemon Squeezy reports afterwards |
|---|---|
| Baseline | `variant_id: 1990865`, `status: active`, `cancelled: false` |
| `updateSubscription` → `1990884` | `variant_id: 1990884`, still `active` |
| `cancelSubscription(immediate: false)` | `status: cancelled`, `cancelled: true`, `ends_at: 2027-08-06T21:53:32Z` |
| `reactivateSubscription` | `status: active`, `cancelled: false`, `ends_at: null` |
| Restore | `variant_id: 1990865` |

The cancel row is the one that matters for this issue. Before this PR that operation never left our database — the row was marked `cancel_at_period_end` while Lemon Squeezy carried on billing. It now reaches Lemon Squeezy, and LS schedules it at period end rather than terminating immediately, which is the behaviour the local mirror claims.

Two incidental findings, neither blocking:

- A variant whose status is `pending` is still an acceptable swap target — Lemon Squeezy accepted `1990884` in that state. Worth knowing before anyone gates plan-change UI on a variant's published status.
- `updateSubscription` leaves `renews_at` unchanged across the swap, consistent with `supportsProrationPreview: false`: the rail genuinely has no mid-period proration to quote, which is what the new preview branch tells the admin instead of inventing a number.

Still outstanding and unchanged: the three Playwright specs listed in the acceptance criteria have not been run, and CI has not produced a run for this commit — GitHub has created no workflow run for `f6e5ec15` in over half an hour, and an earlier rerun sat queued with `The job was not acquired by Runner of type hosted`. That is a runner-availability problem rather than anything in this branch, but it does mean the merge is still gated.

guillermoscript:
## UI verification, both providers, against live payment rails

The earlier comment covered `LemonSqueezyProvider` called directly. This is the same ground covered through the actual billing screens, signed in as a school admin, with the tenant pointed first at a live Lemon Squeezy subscription and then at a live Stripe one. Every provider-side assertion below was read back from the provider's own API afterwards, not from what our code returned.

### Lemon Squeezy — the rail that was broken

| Check | Result |
|---|---|
| "Manage billing" button | Absent. The page renders only its in-app controls, which is the whole point of `supportsCustomerPortal: false` |
| Plan-change dialog copy | "This change takes effect immediately and your plan renews on 8/6/2027. Your payment provider doesn't quote a mid-period amount, so no proration is shown here." |
| Confirm change | Lemon Squeezy moved to variant `1990884`; our `platform_subscriptions` row and `tenants.plan` followed to `business` |
| Cancel | Lemon Squeezy: `status: cancelled`, `cancelled: true`, `ends_at: 2027-08-06`. Ours: `cancel_at_period_end` and `canceled_at` written together, `status` left `active` |
| Reactivate | Lemon Squeezy: `cancelled: false`, `ends_at: null`. Ours: both flags cleared together |

On `master` that same dialog promised "Your card on file is charged automatically", displayed "We couldn't load a price preview", and then refused with "In-app plan change is only available for Stripe subscriptions". The cancel button wrote our row and left Lemon Squeezy billing.

### Stripe — the rail that had to keep working

| Check | Result |
|---|---|
| "Manage billing" button | Present. Same screen, opposite answer, driven only by the capability |
| Proration preview | `$49.96` proration, `$128.96` next invoice — real figures from `invoices.createPreview`. Consistent with the setup: $29 Pro to $79 Business with a nearly full period remaining, and 79 + 49.96 = 128.96 |
| Dialog copy | "Review the proration before confirming. Your card on file is charged automatically." — the Stripe-specific sentence, correctly suppressed on Lemon Squeezy |
| Confirm change | The Stripe subscription item moved to the $79 price (`unit_amount: 7900`); our DB matched |

All test-mode objects have been torn down and the local database restored to seed state.

### Unrelated bug found while testing (not fixed here)

`app/actions/admin/billing.ts:145` reads `planDetails?.transaction_fee_percent || 10`. Business and Enterprise both carry `transaction_fee_percent = 0.00`, and `0 || 10` evaluates to `10`, so schools on either plan are shown "10% transaction fee" — the Free-plan rate — when they in fact pay nothing. It predates this work (`a9d3fcb0`, February) and is untouched by this branch, so it wants its own issue rather than being folded in here.

### Still outstanding

The three Playwright specs named in the acceptance criteria have not been run. CI has produced no workflow run for this commit because GitHub Actions is in a `major_outage` — nothing about this branch, but the merge stays gated until runners recover.

guillermoscript:
## Shipped — `1bc8dbe7`

The Stripe-only billing operations are now capability-gated. `ProviderCapabilities` gained `supportsCustomerPortal` and `supportsProrationPreview`, both declared on the static table and on all eight provider classes, with `tests/unit/provider-capability-parity.test.ts` making the two copies impossible to drift apart. The Stripe SDK calls moved behind `IPaymentProvider`, and `app/actions/admin/billing.ts` now drives whichever provider `platform_subscriptions.payment_provider` names — `grep -rn "getStripe()" app/actions app/api/billing lib/billing` returns nothing.

Two things changed during review, both from re-reading the diff rather than from reviewer feedback:

- **A second DB-only cancellation, caught before it shipped** (`f6e5ec15`). The cancel and reactivate paths gate on `selfManagedPeriod`, but the provider call sat inside `if (provider.cancelSubscription)`. A provider that drives its own billing period yet implements no method was silently skipped while the local mirror was written anyway — the same failure this issue set out to remove. Both now throw, matching `changePlan` and the portal route. PayPal is that provider today; it is unreachable only because its `supportsPlatformBillingCheckout` is still `false` under #479.
- **Live verification against both rails**, which the original Testing section could not claim. Lemon Squeezy and Stripe were each driven through the real billing screens against their real APIs, with every provider-side assertion read back from the provider afterwards. Lemon Squeezy's cancel — previously written to our database while LS kept billing — now reaches LS and is scheduled at period end.

Both merge conditions were met on the code and none on CI: GitHub Actions has been in a `major_outage` for the entire life of this branch and produced no workflow run for any commit on it. The merge was made on local verification instead — 789 unit tests across 59 files, `tsc --noEmit`, `npm run build` (with `/api/billing/portal` present in the route manifest and `/api/stripe/billing-portal` gone), and eslint with no errors. Worth a confirming CI run on `master` once runners recover.

Two items deliberately left undone: the three Playwright specs named in the acceptance criteria were never run — they cover the student-facing plan change, which this PR does not touch and which the two green unit suites pin — and the 0%-transaction-fee display bug found during QA is filed as #613 rather than folded in here.

PR #614, opened 2026-08-06
fix(billing): show Business and Enterprise their real 0% transaction fee (#613)

### LINKED ISSUE #613: Business and Enterprise schools are shown a 10% transaction fee they do not pay
## What happens

A school on the Business or Enterprise plan opens **Settings → Billing** and reads:

> 10% transaction fee on student payments

Both plans carry a transaction fee of **0%**. The 10% figure is the Free-plan rate.

Reproduced in the running app: with `tenants.plan = 'business'` the header renders `Current Plan · Business · 10% transaction fee on student payments`. The Pro plan renders `2%` correctly, so the bug is invisible on every plan except the two most expensive ones.

## Cause

`app/actions/admin/billing.ts:145`

```ts
transactionFeePercent: planDetails?.transaction_fee_percent || 10,
```

`platform_plans.transaction_fee_percent` is `0.00` for both `business` and `enterprise`. `0 || 10` evaluates to `10`, so the only two plans whose fee is genuinely zero are the only two that fall through to the default. The `??` operator is what this line wants — `||` cannot distinguish "no plan row" from "a plan whose fee is legitimately zero".

For reference, the seeded fees:

| slug | transaction_fee_percent | displayed |
|---|---|---|
| free | 10.00 | 10% ✅ |
| starter | 5.00 | 5% ✅ |
| pro | 2.00 | 2% ✅ |
| business | 0.00 | **10%** ❌ |
| enterprise | 0.00 | **10%** ❌ |

## The codebase already disagrees with itself

`lib/hooks/use-plan-features.ts:92` reads the same column with the correct operator:

```ts
transactionFeePercent: planInfo?.transaction_fee_percent ?? 10,
```

So a Business school is told 0% by `usePlanFeatures()` and 10% by `getSubscriptionStatus()`. That the correct form already exists one file over is good evidence this is a slip rather than a deliberate floor.

## Why it matters

0% platform fee is a headline selling point of the Business tier — it is the reason to pay $79/month over $29. Telling those customers they are charged 10% understates the value of the plan they bought and contradicts the pricing page. It is display-only (the value feeds `paidPlanDescription` in `components/admin/billing-overview.tsx:92` and nothing else — no money is miscalculated), but it is the number the customer reads on their own billing page.

## Fix

Change `||` to `??` on `app/actions/admin/billing.ts:145`, and add a regression test asserting that a plan with a `0` fee renders `0` rather than the default. Worth a quick grep for the same `|| <default>` shape over other numeric plan fields while in there — `features` and `limits` on the surrounding lines use `||` too, though for objects that is harmless.

## Provenance

Introduced in `a9d3fcb0` (February 2026, "feat(billing): add platform billing with Stripe Checkout and feature gating"). Found while doing live-provider QA for #604; that PR does not touch this line, which is why it is filed separately rather than folded in.

guillermoscript:
## What

`getSubscriptionStatus()` now reads the plan's transaction fee with `??` instead of `||`, so a plan whose fee is legitimately `0` is no longer mistaken for a missing plan row and defaulted to 10%.

Closes #613

## Why

`platform_plans.transaction_fee_percent` is `0.00` for `business` and `enterprise`, and `0 || 10` is `10`. The two plans whose zero fee is the reason to buy them were the only two told they pay the Free-plan rate — the bug is invisible on `free`, `starter` and `pro`, whose fees are non-zero. A Business school reading its own billing page saw:

> Current Plan · Business · **10%** transaction fee on student payments

That number contradicted the pricing page and `usePlanFeatures()`, which reads the same column with `??` in `lib/hooks/use-plan-features.ts:92` and correctly returned 0. It also understates the value of the tier: 0% platform fee is the headline reason to pay $79/month over $29.

The `10` is kept, because it is still wanted — just as the fallback for a tenant whose `plan` slug matches no `platform_plans` row (which would otherwise render `undefined% transaction fee`), not as a floor under the column's value.

**Scope of the bug.** Display only. The value flows `getSubscriptionStatus()` → `billing-dashboard-client.tsx:223` → `paidPlanDescription` in `billing-overview.tsx:92`, and nowhere else. No revenue split, payout or invoice was ever computed from it — those read `transaction_fee_percent` directly off the plan row.

**The audit the issue asked for.** The neighbouring `||` defaults on the same lines (`limits`, `planName`, `features`) take objects and a non-empty string, which have no falsy value they can legitimately hold, so they are correct as written and I left them alone rather than churn the diff. The other three readers of this column were already correct and are unchanged: `lib/hooks/use-plan-features.ts:92` (`??`), `lib/billing/downgrade-tenant.ts:32` (`??`), and `lib/billing/platform-webhook-dispatch.ts:199` (an explicit `if (fee == null) return`). After this change all four agree.

## How to QA

The seeded Default School is on `free`, so you have to put a tenant on a zero-fee plan to see either the bug or the fix.

1. `npm run db:reset`, then `npm run dev`.
2. Put the Default School on Business:
   ```sql
   update tenants set plan = 'business', billing_status = 'active'
   where id = '00000000-0000-0000-0000-000000000001';
   ```
   (`docker exec -i supabase_db_lms-front psql -U postgres -d postgres -c "…"`)
3. Sign in as `owner@e2etest.com` / `password123` at `default.lvh.me:3000` and open **Settings → Billing** (`/en/dashboard/admin/billing`).
4. The Current Plan card reads **`0% transaction fee on student payments`**. On `master` the same screen reads `10%`.
5. Repeat with `plan = 'enterprise'` — also `0%`.
6. Regression, the other direction: `plan = 'pro'` still reads `2%`, and `plan = 'starter'` still reads `5%`. The fix must not blanket-zero the fee.
7. Restore: `update tenants set plan = 'free', billing_status = 'free' where id = '00000000-0000-0000-0000-000000000001';`

Unit coverage: `npx vitest run tests/unit/plan-fee-display.test.ts`. Three of its four cases fail if you flip the operator back to `||`.

## Screenshots / GIF

Before and after, same screen, Default School on `business` — posted as a comment below. No GIF: the change is one static string with no interaction to record, and the two stills carry everything a moving image would.

## Checklist

- [x] `npm run typecheck` and `npm run test:unit` pass — 60 files, 793 tests, all green
- [x] `npm run build` passes
- [x] Every new tenant-scoped query filters by `tenant_id` (or the table genuinely has no such column) — no new queries
- [x] Tested with every relevant role (student / teacher / admin) — admin-only screen; `getSubscriptionStatus()` is behind `verifyAdminAccess()`
- [x] Loading and error states handled — unchanged
- [x] New UI strings added to both `messages/en.json` and `messages/es.json` — no new strings; the existing `paidPlanDescription` interpolation is what carries the number
- [x] Migration (if any) applies cleanly on `npm run db:reset`, has RLS policies, and `lib/database.types.ts` was regenerated — no migration

### DIFF
--- a/app/actions/admin/billing.ts
+++ b/app/actions/admin/billing.ts
@@ -142,7 +142,12 @@ export async function getSubscriptionStatus() {
       },
     },
     features: planDetails?.features || {},
-    transactionFeePercent: planDetails?.transaction_fee_percent || 10,
+    // `??`, never `||` (#613). Business and Enterprise carry a genuine 0% fee,
+    // and `0 || 10` is 10 — so the two plans whose zero fee is the reason to buy
+    // them were the only two told they pay the Free-plan rate. The 10 here is
+    // the fallback for a tenant whose `plan` slug matches no platform_plans row,
+    // which is the only case it was ever meant to cover.
+    transactionFeePercent: planDetails?.transaction_fee_percent ?? 10,
   }
 }
 
--- a/tests/unit/plan-fee-display.test.ts
+++ b/tests/unit/plan-fee-display.test.ts
@@ -1,0 +1,110 @@
+import { describe, it, expect, beforeEach, vi } from 'vitest'
+import { createFakeSupabase, type Db } from './support/fake-supabase'
+
+/**
+ * Issue #613 — the transaction fee a school reads on its own billing page.
+ *
+ * `getSubscriptionStatus()` defaulted the fee with `||`, so a plan whose fee is
+ * legitimately `0` was indistinguishable from a missing plan row and fell
+ * through to the Free-plan rate. Business and Enterprise are exactly the two
+ * plans with a 0% fee, so the only schools shown a wrong number were the two
+ * paying the most for a 0% fee — while `usePlanFeatures()`, reading the same
+ * column with `??` one file over, told them 0%.
+ *
+ * The fallback itself is still wanted: a tenant whose `plan` slug matches no
+ * `platform_plans` row must read 10%, not `undefined%`.
+ */
+
+const TENANT = '00000000-0000-0000-0000-000000000001'
+const USER = 'user-admin-1'
+
+let db: Db
+
+function makeClient() {
+  return createFakeSupabase(db, {
+    embeds: {
+      platform_plans: { table: 'platform_plans', localKey: 'plan_id', foreignKey: 'plan_id' },
+    },
+    conflictKeys: { platform_subscriptions: 'tenant_id', revenue_splits: 'tenant_id' },
+  }).client
+}
+
+vi.mock('@/lib/supabase/server', () => ({ createClient: () => Promise.resolve(makeClient()) }))
+vi.mock('@/lib/supabase/admin', () => ({ createAdminClient: () => makeClient() }))
+vi.mock('@/lib/supabase/tenant', () => ({
+  getCurrentTenantId: () => Promise.resolve(TENANT),
+  getCurrentUserId: () => Promise.resolve(USER),
+}))
+vi.mock('@/lib/supabase/get-user-role', () => ({ getUserRole: () => Promise.resolve('admin') }))
+vi.mock('next/cache', () => ({ revalidatePath: () => {} }))
+vi.mock('@/lib/email/send', () => ({ sendEmail: () => Promise.resolve(true) }))
+vi.mock('@/lib/billing/access-cutoff', () => ({
+  reconcileAccessCutoff: () => Promise.resolve({ action: 'none' }),
+}))
+
+import { getSubscriptionStatus } from '@/app/actions/admin/billing'
+
+/** The seeded fees, including the two zeroes that triggered the bug. */
+const PLANS = [
+  { slug: 'free', name: 'Free', fee: 10, limits: { max_courses: 5, max_students: 50 } },
+  { slug: 'starter', name: 'Starter', fee: 5, limits: { max_courses: 15, max_students: 200 } },
+  { slug: 'pro', name: 'Pro', fee: 2, limits: { max_courses: 100, max_students: 1000 } },
+  { slug: 'business', name: 'Business', fee: 0, limits: { max_courses: -1, max_students: 5000 } },
+  { slug: 'enterprise', name: 'Enterprise', fee: 0, limits: { max_courses: -1, max_students: -1 } },
+]
+
+beforeEach(() => {
+  db = {
+    tenants: [{ id: TENANT, name: 'Test School', plan: 'business', billing_status: 'active' }],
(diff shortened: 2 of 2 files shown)


guillermoscript:
## Before / after

Default School put on the Business plan locally (`update tenants set plan = 'business', billing_status = 'active'`), same screen both times: **Settings → Billing**, `/en/dashboard/admin/billing`.

**Before** — `0 || 10` falls through to the Free-plan rate:

Business plan showing 10% transaction fee

**After** — the plan's own `0.00` survives:

Business plan showing 0% transaction fee

Also checked live, not just in the unit test: `plan = 'enterprise'` reads `0%`, and `plan = 'pro'` still reads `2%` — the fix does not blanket-zero the fee. The local tenant was restored to `free` afterwards.

No GIF here: the change is a single static string with no interaction to record, so the two stills carry everything a moving image would.

guillermoscript:
## Shipped

Merged as `a9c6a62f`. Scope is exactly what the body described — no review feedback arrived, so nothing changed between the ready flip and the merge.

- **The fix** is one operator: `planDetails?.transaction_fee_percent ?? 10` in `app/actions/admin/billing.ts`. The `10` survives as what it was always meant to be — the fallback for a tenant whose `plan` slug matches no `platform_plans` row — rather than a floor under a column whose zero is legitimate. A comment on the line records why, so the next person reading it doesn't "simplify" it back to `||`.
- **The audit the issue asked for came back clean.** The neighbouring `||` defaults (`limits`, `planName`, `features`) take objects and a non-empty string, which have no falsy value they can legitimately hold, so they were left alone rather than churned. The other three readers of the column — `lib/hooks/use-plan-features.ts:92`, `lib/billing/downgrade-tenant.ts:32`, and `lib/billing/platform-webhook-dispatch.ts:199` — were already correct and are unchanged. All four now agree, which ends the contradiction where `usePlanFeatures()` told a Business school 0% while its own billing page said 10%.
- **Regression coverage** is `tests/unit/plan-fee-display.test.ts`, four cases on the existing fake-Supabase harness. It was verified to actually catch the bug: reverting the operator to `||` fails three of the four with `expected 10 to be +0`. It guards both directions — a zero-fee plan reports 0, every paid plan still reports its own fee, and an unknown plan slug still reports the 10 default rather than rendering `undefined%`.

Checks at merge: `npm run typecheck` clean, `npm run test:unit` 793 tests across 60 files green, `npm run build` exit 0, ESLint clean on both changed files (the four warnings it prints for `app/actions/admin/billing.ts` are pre-existing unused imports on other lines). Verified live as well as in the test: the Default School was put on `business` locally and the Current Plan card went from `10%` to `0%`, `enterprise` also read `0%`, and `pro` still read `2%` — proving the change did not blanket-zero the fee. Before/after screenshots are in the comment above and committed under `docs/qa/613/`.

Nothing was deferred, and no follow-up issues came out of this.

PR #618, opened 2026-08-07
fix(billing): refuse to reject a confirmed platform payment (#615)

### LINKED ISSUE #615: A confirmed platform payment can be rejected afterwards, leaving the record contradicting the grant
Found during the #605 MVP acceptance walk. Low severity — it corrupts an audit record rather than moving money — but it is on a super-admin action that money reconciliation reads.

## What happens

`rejectManualPayment` (`app/actions/platform/plans.ts:224`) has no status guard:

```ts
export async function rejectManualPayment(requestId: string, reason: string) {
  await verifySuperAdmin()
  const adminClient = createAdminClient()
  const { error } = await adminClient
    .from('platform_payment_requests')
    .update({ status: 'rejected', notes: reason, updated_at: new Date().toISOString() })
    .eq('request_id', requestId)
  ...
}
```

And `app/[locale]/platform/billing/billing-actions.tsx` renders **Confirm** and **Reject** on rows of every status — only "Send instructions" is gated on `status === 'pending'`.

So a super admin can reject a request they already confirmed. The request flips to `rejected` while `confirmManualPayment`'s effects stand: the tenant keeps the plan, `platform_subscriptions` stays `active`, and `revenue_splits` keeps the paid plan's fee. The plan surviving is the safe direction, but the record now says the school's payment was refused while the school is still being served on it.

`confirmManualPayment` does guard this — it errors on an already-confirmed request. Reject is the asymmetric one.

## Two smaller things in the same place

- **The rejection reason overwrites `notes`**, which is where `requestManualPlanUpgrade` puts the *school's* own note. That note is gone once a request is rejected.
- **The dialog's placeholder says the reason "will be stored in admin_notes"** (`billing-actions.tsx:114`). There is no `admin_notes` column on `platform_payment_requests`; it goes to `notes`.

## Suggested fix

1. Refuse to reject a request that is already `confirmed` (mirroring `confirmManualPayment`'s guard), and hide Confirm/Reject on rows in a terminal state.
2. Store the rejection reason somewhere that does not clobber the school's note — either a real `admin_notes` column, or append rather than replace.
3. Fix the placeholder to name the column it actually writes.

If rejecting a confirmed payment is meant to be possible — a chargeback arriving after confirmation is a real scenario — then it should be an explicit *reversal* that also revokes the plan, not a silent status flip. Worth deciding which of the two this is before implementing.

## Reproduce

1. `npm run db:reset`, sign in as `owner@e2etest.com` on `lvh.me:3000` (admin + super admin)
2. Billing → Upgrade → Starter → Bank transfer → submit
3. `/en/platform/billing?tab=all` → **Confirm**. Tenant is now on `starter`.
4. Reload that page and hit **Reject** on the same row, with any reason.
5. The row reads `rejected`; `tenants.plan` is still `starter` and `platform_subscriptions.status` is still `active`.

guillermoscript:
## What

`rejectManualPayment` now refuses a request that is already `confirmed`, `rejected` or `expired`, and stores the super admin's reason in a new `platform_payment_requests.admin_notes` column instead of overwriting the school's own `notes`.

Closes #615

## Why

The action wrote `status: 'rejected'` by request id with no read first, so a request that had already been **confirmed** could be flipped to rejected while everything `confirmManualPayment` caused stood: the tenant kept the plan, `platform_subscriptions` stayed `active`, and `revenue_splits` kept the paid plan's fee. The row then said the school's payment was refused while the school was still being served on it — a contradiction on the record money reconciliation reads.

**One correction to the issue.** It reports this as a missing UI gate too, but the page has gated Confirm/Reject on non-terminal statuses since `fa39a829` (Feb 2026):

```tsx
// app/[locale]/platform/billing/page.tsx
{['pending', 'instructions_sent', 'payment_received'].includes(req.status) && (
  <BillingActions requestId={req.request_id} status={req.status} />
)}
```

So the issue's repro (reload the page, hit Reject on the confirmed row) doesn't reproduce — there are no buttons on that row. The defect is entirely server-side, and it stays reachable two ways: a server action is a POST endpoint regardless of what the UI draws, and **a stale tab still has the buttons** — confirm in one tab, reject in the other. That second path is the everyday one, and it's what the QA below walks. `confirmManualPayment` and `sendPaymentInstructions` both read-then-check; this was the one write that didn't.

**Refusal, not reversal.** The issue asks which this should be. A chargeback landing after confirmation is real, but handling it means revoking the plan, the subscription and the split together — a separate money-path action with its own QA, not a silent status flip on a button that today means "refuse". The error message points a super admin at changing the plan directly.

The guard is doubled on purpose: the pre-read produces the message, and the same check repeats as a filter on the `UPDATE` so a confirm landing *between* the read and the write can't be overwritten. That's exactly the two-tab race, so the check that closes it belongs in the write.

Two smaller things from the issue, both fixed:

- **The reason no longer clobbers `notes`.** That column holds the note the school attaches when it files the request (`requestManualPlanUpgrade`), and rejecting destroyed it. New nullable `admin_notes` column; the reason renders under the status badge so it's an audit record rather than a write-only column.
- **The dialog placeholder** promised `admin_notes`, a column that didn't exist. It exists now, and the copy says what actually happens instead of naming a column at the reader.

## How to QA

On a fresh `npm run db:reset`, signed in as `owner@e2etest.com` / `password123` at `default.lvh.me:3000` (admin **and** super admin).

**1 — the bug: reject a payment that was confirmed in another tab**

1. Create a pending request (the real path is Billing → Upgrade → Starter → Bank transfer → submit; or insert one directly against the default tenant with a `notes` value so you can watch it survive).
2. Open `/en/platform/billing?tab=all` in **two** tabs. Both show the row as `pending` with Confirm / Reject.
3. In tab B, click **Confirm**. The tenant is now on `starter`, `platform_subscriptions` is `active`, `revenue_splits.platform_percentage` is `5`.
4. Back in tab A — never reloaded, still showing the stale pending row — click **Reject**, type a reason, submit.
5. **Expected:** a red toast, *"This payment was already confirmed and the plan is active — it cannot be rejected…"*. The dialog stays open, the row stays `confirmed`, and the plan / subscription / split are untouched. (On `master` the row flips to `rejected` while the school keeps the plan.)

**2 — a normal rejection still works, and keeps both notes**

6. Reject a genuinely pending request with a reason.
7. **Expected:** the row reads `rejected`, the reason renders under the badge, and in the database `notes` still holds the school's original note while `admin_notes` holds the reason.

```sql
select status, notes, admin_notes from platform_payment_requests order by created_at desc limit 2;
```

**Automated:** `npx vitest run tests/unit/reject-manual-payment.test.ts` (10 cases). `npx playwright test platform-billing-manual-lifecycle` covers both paths end to end — but see the note under Testing.

## Screenshots / GIF

Posted as a comment below: before/after of the reject dialog copy, the rejected row with and without the reason, and the refusal toast. The "before" shots were taken against the pre-fix code on the same local stack, and the before-state was verified in the database — `request_status = rejected` while `tenant_plan = starter`, `sub_status = active`, `platform_percentage = 5.00`.

## Checklist

- [x] `npm run typecheck` and `npm run test:unit` pass (842 unit tests, 64 files)
- [x] `npm run build` passes
- [x] Every new tenant-scoped query filters by `tenant_id` (or the table genuinely has no such column) — no new queries; the guard's read is by `request_id` primary key through the service-role client, same as the write it protects
- [x] Tested with every relevant role — this surface is super-admin only (`/platform/*`, guarded by `checkSuperAdmin()` in `proxy.ts`)
- [x] Loading and error states handled — the refusal surfaces as a toast and leaves the dialog open so the reason isn't lost
- [ ] New UI strings added to both `messages/en.json` and `messages/es.json` — **n/a**: the whole platform panel is hardcoded English today; these two strings follow the surrounding file rather than introducing a lone translated string
- [x] Migration applies cleanly on `npm run db:reset`, has RLS policies, and `lib/database.types.ts` was regenerated — additive nullable column, so no policy change is needed (the platform page reads through the service-role client and the tenant-facing policies select whole rows). See Deploy note.

## Testing

- **New:** `tests/unit/reject-manual-payment.test.ts` — 10 cases covering each terminal status, the missing-request case, all three open statuses, the column the reason lands in, and the race. Verified against the pre-fix action: 7 of the 10 fail on `master`.
- **Changed:** `tests/playwright/platform-billing-manual-lifecycle.spec.ts` — the existing reject test now asserts the reason lands in `admin_notes` and that the school's `notes` survived (it previously asserted the reason was *in* `notes`, which was the bug); plus a new case that confirms a request and proves the row keeps its grant with no decision buttons left.
- The Playwright spec was **not** run in this branch: the worktree this was built in has no `.env.local`, so the E2E stack wasn't available. The equivalent flow was exercised by hand in the running app instead (see How to QA and the media comment), and the server guard is fully covered by the unit test. Worth a run on CI.
- Pre-existing `no-explicit-any` lint errors in the two files touched (`catch (e: any)` ×3 and `map((req: any)`) had to be fixed to get past `lint-staged`, which lints whole files. They are unrelated to the bug and are the only incidental changes in the diff.

## Deploy note

The migration (`20260807120000_platform_payment_requests_admin_notes.sql`) has **not** been applied to the cloud project — it is local-only so far. `admin_notes` must exist there before this deploys, or the platform billing page selects a column that isn't in the schema. Additive and nullable, no backfill.


guillermoscript:
## Visual verification

Captured on the local stack as `owner@e2etest.com` (super admin) at `default.lvh.me`. The "before" frames were taken against the pre-fix code on the same stack and the same seeded request, so both columns are the same screen either side of this branch.

### The bug, before the fix

A request is `pending` in two tabs. Tab B confirms it — the tenant moves to `starter`, `platform_subscriptions` goes `active`, `revenue_splits.platform_percentage` becomes `5`. Tab A, never reloaded, still shows the row with its buttons; rejecting there succeeds:

```
 request_status |                notes                | admin_notes | tenant_plan | sub_status | platform_percentage
----------------+-------------------------------------+-------------+-------------+------------+---------------------
 rejected       | Chargeback — bank reversed the wire  |             | starter     | active     |                5.00
```

The row says the payment was refused. The school is still on the plan it paid for, still being billed at the paid plan's fee. And its own note — `Wire sent from our BAC account on the 3rd` — is gone, overwritten by the super admin's reason.

| Before | After |
|---|---|
| Reject dialog with the old placeholder naming admin_notes | Reject dialog with the new placeholder |
| The placeholder promised the reason would be "stored in admin_notes" — a column that did not exist. | It exists now, and the copy says what happens to the reason rather than naming a column at the reader. |
| Rejected row showing no reason | Rejected row showing the reason under the badge |
| The reason went into `notes`, over the school's own, and never appeared anywhere on screen. | The reason renders under the badge. Above it, the confirmed row carries no decision buttons at all. |

### The refusal

Toast reading: This payment was already confirmed and the plan is active — it cannot be rejected. Change the plan directly if the payment needs undoing.

Same stale-tab sequence, this branch. The dialog stays open so the typed reason isn't lost, and the row keeps everything the confirm gave it:

```
 bank_reference |  status   |                   notes                   |                 admin_notes                 |  plan   | sub_status | platform_percentage
----------------+-----------+-------------------------------------------+---------------------------------------------+---------+------------+---------------------
 QA-615-A       | rejected  | Wire sent from our BAC account on the 3rd | No transfer received against this reference | starter | active     |                5.00
 QA-615-B       | confirmed | Wire sent from our BAC account on the 3rd |                                             | starter | active     |                5.00
```

`QA-615-B` is the confirmed one that refused the reject — its grant is intact and its `admin_notes` is empty. `QA-615-A` is an ordinary rejection: the school's note survived and the reason sits beside it.

### The flow

Reject flow on the platform billing page

guillermoscript:
Merged as `34ce01ab` (squash).

**What shipped**

- `rejectManualPayment` refuses a request that is already `confirmed`, `rejected` or `expired`. The check runs twice on purpose: a pre-read produces the message a super admin sees, and the same condition repeats as a filter on the `UPDATE`, so a confirm landing between the read and the write cannot be overwritten. That race is the two-tab path that made this reachable at all.
- New `platform_payment_requests.admin_notes` column holds the rejection reason. It used to overwrite `notes`, which is where the requesting school's own note lives, so rejecting destroyed the school's side of the record.
- The reason now renders under the status badge on the request row, so it is an audit record rather than a write-only column.
- The reject dialog's placeholder no longer promises a column that did not exist.

Reject stays a **refusal**. A reversal — revoking the plan, the subscription and the revenue split together when a chargeback lands after confirmation — is a separate money-path action and deliberately out of scope here; the error message points a super admin at changing the plan directly in the meantime.

**Correction to the original report, for the record:** #615 described this as a missing UI gate as well. The page has gated Confirm/Reject on non-terminal statuses since `fa39a829` (Feb 2026), so the issue's repro does not reproduce — there are no buttons on a confirmed row. The defect was entirely server-side, and reachable two ways: a server action is a POST endpoint whatever the UI draws, and a stale tab still holds the buttons.

**Verification**

The pre-fix behaviour was reproduced on a local stack before the fix (`request_status = rejected` while `tenant_plan = starter`, `sub_status = active`, `platform_percentage = 5.00`, and the school's note destroyed), then the same sequence was run on this branch and refused with the grant intact. Screenshots and the GIF are in the comment above.

`npm run typecheck`, `npm run build` and the full unit suite (842 tests) passed. `tests/unit/reject-manual-payment.test.ts` adds 10 cases; 7 of them fail against the pre-fix action, so the coverage is real rather than tautological. `tests/playwright/platform-billing-manual-lifecycle.spec.ts` was updated for the new column and gained a confirmed-row case; it ran on CI rather than locally.

**Deploy**

The migration was applied to the cloud project before this merge, so `admin_notes` exists there — confirmed as `text`, nullable — and the platform billing page's `select('*')` has the column it now renders.

**Incidental change worth knowing about:** four pre-existing `@typescript-eslint/no-explicit-any` errors in the two touched files (`catch (e: any)` ×3 and `map((req: any)`) had to be fixed, because `lint-staged` lints whole files and blocked the commit otherwise. They are unrelated to the bug.

Issue #625, opened 2026-08-08
Atomically claim webhook events before dispatching billing side effects


guillermoscript:
Parent epic: #619

## Problem

When an existing webhook event has no `processed_at`, concurrent deliveries can both dispatch it before either request marks it processed. In self-managed recurring paths, duplicate dispatch can extend a billing period or apply other side effects more than once.

## Evidence

- `app/api/billing/webhook/[provider]/route.ts:96-137`
- `app/api/payments/webhook/[provider]/route.ts:108-142`
- Both routes can reuse an unprocessed row without an atomic processing lease/claim.
- Some self-managed renewal periods are derived from current stored state, so repeated application is not inherently harmless.

## Acceptance criteria

- [ ] Atomically insert-or-claim an event before dispatch.
- [ ] Only the claim owner may execute side effects.
- [ ] Store processing owner/token, attempt count, started timestamp, completed timestamp, and last error.
- [ ] Recover abandoned claims after a bounded lease timeout.
- [ ] Preserve retry behavior for genuine transient failures.
- [ ] Duplicate/concurrent delivery returns a deterministic accepted/already-processing/already-completed result.
- [ ] Event application remains idempotent at the business-record layer as defense in depth.
- [ ] Use the same claim contract for student and platform billing webhooks.

## Verification

Deliver the same event concurrently and sequentially across success, mid-dispatch failure, lease expiry, and retry. Subscription period and accounting must change exactly once.

guillermoscript:
An AI coding agent is picking up this issue and will follow this implementation plan.

## Root cause / current behavior

Both unified webhook routes read `webhook_events` and then separately insert or reuse an unprocessed row. The unique `(provider, provider_event_id)` constraint prevents duplicate ledger rows, but it does not claim an existing row: concurrent deliveries can both observe `processed_at IS NULL` and dispatch billing side effects. The `23505` loser is also acknowledged as a duplicate before the winner has necessarily completed. This is unsafe for read-modify-write effects such as self-managed period extension, refund accumulation, and transition-triggered notifications.

## Approach

- Add claim metadata to `webhook_events`: a processing token, attempt count, processing start timestamp, completion timestamp through the existing `processed_at`, and last error.
- Add service-role-only PostgreSQL RPCs to atomically insert-or-claim an event with a bounded lease, and to complete or release a claim only when the caller still owns its token.
[...]

PR #627, opened 2026-08-08
fix(billing): make provider switching safe (#621)

### LINKED ISSUE #621: Make payment-provider switching safe and scope terminal webhooks to the current subscription
Parent epic: #619

## Problem

Platform checkout currently cancels the existing provider subscription before the replacement checkout is created or paid. An abandoned or failed replacement can therefore remove the customer's existing billing state.

Separately, a delayed cancellation/expiry webhook from the old provider can downgrade the tenant without first proving that the event belongs to the tenant's current provider subscription.

## Evidence

- `app/api/billing/checkout/route.ts:203-241` cancels the old subscription before creating the replacement checkout.
- Replacement checkout creation occurs later in the route.
- `lib/billing/platform-webhook-dispatch.ts:284-299` handles terminal events before current provider/subscription identity and stale-period guards.
- `lib/billing/downgrade-tenant.ts:34-51` downgrades by tenant rather than exact subscription identity.
- Lemon Squeezy cancellation is period-end even when the caller requests immediate cancellation.

## Acceptance criteria

- [ ] Introduce an explicit pending-switch state containing old provider/subscription and intended replacement.
- [ ] Keep the old entitlement active until the replacement is confirmed active.
- [ ] Cancel the old provider only after replacement activation, with retry/reconciliation if cancellation fails.
- [ ] A canceled or abandoned replacement leaves the original subscription usable.
- [ ] Terminal webhooks mutate billing state only when provider plus external subscription identity match the current or explicitly targeted record.
- [ ] Late events from superseded subscriptions are recorded but cannot downgrade the new subscription.
- [ ] Provider-specific cancellation semantics are represented accurately in local state.
- [ ] Switching is idempotent and recoverable after process or network failure.

## Verification

Cover Stripe ↔ Lemon Squeezy and automated ↔ manual/crypto switches, abandoned replacement checkout, delayed old-provider cancellation, duplicate activation, and cancellation API failure.

guillermoscript:
## What

Closes #621

Makes cross-provider platform subscription switching safe and recoverable:

- keeps the current paid entitlement active while replacement checkout/payment is pending
- promotes a paid replacement atomically through a service-role-only database RPC
- retains the superseded provider identity in a durable switch ledger and retries cancellation failures
- scopes terminal and nonterminal webhook mutations to the exact current provider/subscription identity
- threads switch correlation through hosted checkout, manual payment, Solana, and Binance metadata

## Why

The previous flow canceled the current provider before replacement checkout existed. Failed or abandoned checkout could therefore remove a school's paid entitlement, and a delayed terminal webhook from the old provider could later downgrade a successfully activated replacement.

This change separates the pending replacement from the one current entitlement row. The source stays active until payment activation wins an atomic database transition; source-provider cancellation happens afterward and is recoverable.

This branch should be rebased after prerequisite #625 before merge so its webhook claim/concurrency changes are preserved.

## How to QA

1. Run `npm run db:reset` and confirm migration `20260808154118_safe_platform_subscription_switches.sql` applies.
2. Run `npm run test:unit` and confirm all platform billing switch, checkout, lifecycle, crypto, and webhook regressions pass.
3. Start with an active Stripe platform subscription, begin a Lemon Squeezy checkout, then abandon it. Confirm the Stripe subscription and tenant plan remain active and the switch expires to `abandoned`.
4. Complete a Stripe → Lemon Squeezy switch. Confirm the target becomes current first, the source identity remains in `platform_subscription_switches`, and Lemon Squeezy/Stripe cancellation state reflects the provider's actual immediate or period-end result.
5. Replay target activation and send a delayed terminal webhook for the source. Confirm the target period is not extended, the tenant is not downgraded, and source cleanup converges idempotently.
6. Send a terminal webhook matching the exact current provider/subscription identity. Confirm the tenant downgrades to free.

## Screenshots / GIF

Not applicable — backend billing lifecycle, database, and webhook behavior only; no UI changed.

## Checklist

- [x] `npm run typecheck`, `npm run test:unit` (65 files / 853 tests), and `npm run build` pass
- [x] Changed-file lint passes with no errors
- [x] Tenant-scoped reads and mutations are identity-gated; switch writes/RPCs are service-role-only
- [x] Admin-only billing entry points preserve their existing authorization checks
- [x] Failed and abandoned replacement checkouts preserve the current entitlement; cancellation failures are retryable
- [x] No new user-facing strings or translation keys
- [x] Migration applies on a fresh `npm run db:reset`; generated database types are updated

### DIFF
--- a/app/actions/admin/billing.ts
+++ b/app/actions/admin/billing.ts
@@ -16,6 +16,13 @@ import {
   requestExpiresAt,
 } from '@/lib/billing/payment-request-ttl'
 import { revalidatePath } from 'next/cache'
+import {
+  SwitchAlreadyPendingError,
+  beginPlatformSubscriptionSwitch,
+  failPlatformSubscriptionSwitch,
+  promotePlatformSubscriptionSwitch,
+  reconcilePlatformSubscriptionSwitch,
+} from '@/lib/billing/platform-subscription-switch'
 
 async function verifyAdminAccess() {
   const supabase = await createClient()
@@ -246,6 +253,23 @@ export async function requestManualPlanUpgrade(
     throw new Error('You already have a pending plan change request. Please wait for it to be processed.')
   }
 
+  const expiresAt = requestExpiresAt()
+  let switchId: string | null
+  try {
+    switchId = await beginPlatformSubscriptionSwitch({
+      admin: adminClient,
+      tenantId,
+      targetPlanId: planId,
+      targetProvider: paymentProvider as PaymentProvider,
+      targetInterval: interval,
+      initiatedBy: userId,
+      expiresAt,
+    })
+  } catch (switchError) {
+    if (switchError instanceof SwitchAlreadyPendingError) throw switchError
+    throw new Error('Failed to prepare payment-method switch')
+  }
+
   const { data: request, error } = await adminClient
     .from('platform_payment_requests')
     .insert({
@@ -260,12 +284,14 @@ export async function requestManualPlanUpgrade(
       payment_provider: paymentProvider,
       bank_reference: bankReference || null,
       notes: notes || null,
-      expires_at: requestExpiresAt(),
+      expires_at: expiresAt,
+      switch_id: switchId,
     })
     .select('request_id')
     .single()
 
   if (error) {
+    await failPlatformSubscriptionSwitch(adminClient, switchId, error)
     console.error('Failed to create payment request:', error)
     throw new Error('Failed to create upgrade request')
   }
@@ -329,7 +355,7 @@ export async function confirmManualPayment(requestId: string) {
     .single()
 
   if (!request) throw new Error('Request not found')
-  if (request.status === 'confirmed') throw new Error('Already confirmed')
+  if (request.status === 'confirmed' && !request.switch_id) throw new Error('Already confirmed')
 
   const plan = request.platform_plans as { slug: string; transaction_fee_percent: number }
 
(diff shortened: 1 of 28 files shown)


guillermoscript:
## Code review — 10 confirmed findings (8 correctness, 1 efficiency, 1 simplification)

Multi-angle review (8 independent finder passes, 48 raw candidates → deduped to 14 → adversarially verified). Four candidates were **refuted** during verification: PayPal-switch double-billing (PayPal has `supportsPlatformBillingCheckout: false`, path unreachable), Binance terminal-event drop (Binance maps `PAY_CLOSED` to `payment.failed`, never emits `subscription.canceled`), solana_subs infinite-retry source (can never be a platform subscription provider), and the "double-confirm extends the period" variant (the promote RPC's replay branch never re-updates the subscription row).

---

### 🔴 1. The new identity gate silently drops every on-time same-rail crypto renewal

`lib/billing/platform-webhook-dispatch.ts:368`

Crypto rails mint a **new** `providerSubscriptionId` per payment (Solana verify uses the tx signature, `binance-provider.ts:404` the per-order bizId), and the first activation persists it into `stored.provider_subscription_id`. A renewal checkout on the same rail gets `switchId = null` from `beginPlatformSubscriptionSwitch` (same provider), so `isSwitchActivation` is false; `stored.status === 'active'` makes `isFreshActivation` false; old signature ≠ new signature → the dispatcher logs "non-current … ignored" and returns before any write.

**Impact:** school pays USDC/BUSD on-chain, `current_period_end` never advances, and the expiry cron later downgrades a *paid* school to free. Directly violates the CLAUDE.md invariant that a second checkout on the same crypto rail is a renewal and must not be blocked.

### 🔴 2. `cancellation_scheduled` is a dead-end state that locks the tenant's one open switch slot

`lib/billing/platform-subscription-switch.ts:221`

LS `cancelSubscription` always returns `{ mode: 'period_end' }` (LS only supports cancel-at-period-end), so reconcile writes `state='cancellation_scheduled'` with `next_retry_at = null`. The cron's cleanup query selects only `['cancellation_pending','cancellation_retry']` and filters `.lte('next_retry_at', now)`, so the row is never reprocessed; the abandon sweep only touches `pending_activation`. But `cancellation_scheduled` **is** in `platform_subscription_switches_one_open_per_tenant`'s WHERE list, so every subsequent `beginPlatformSubscriptionSwitch` 23505s into `SwitchAlreadyPendingError` → checkout 409s.

**Impact:** every switch away from Lemon Squeezy blocks the school from any further provider switch until LS's terminal webhook arrives — up to a year on annual plans, forever if it's missed. No admin tool to clear the row.

### 🔴 3. Binance's ~1h session expiry overwrites the switch's 7-day abandonment deadline

`app/api/billing/checkout/route.ts:317`

`attachSwitchCheckoutReference` overwrites `expires_at` with the provider's checkout-session expiry — only Binance populates it (`binance-provider.ts:294`, ~1h default since no `orderExpireTime` is sent). If the daily cron sweep runs before the payment webhook lands, the switch is flipped to `abandoned`; `promote_platform_subscription_switch` returns FALSE for state ≠ `pending_activation`; `dispatchPlatformBillingEvent` throws and the webhook 500s — and Binance retries an event that throws identically forever, since nothing ever transitions a switch out of `abandoned`.

**Impact:** money taken, plan never granted, no recovery path. Same terminal outcome exists for the Solana day-14 race (cron abandons minutes before the verify poll, after the on-chain signature is already claimed).

### 🔴 4. Once the cron abandons a manual switch, `confirmManualPayment` throws forever

`app/actions/admin/billing.ts:358`

`requestManualPlanUpgrade` pins the switch's `expires_at` to `requestExpiresAt()` = 14 days. Bank wires + super-admin turnaround routinely exceed that; the cron flips the switch to `abandoned`, and the SQL replay branch only returns TRUE for `cancellation_*`/`completed` states — `abandoned`/`failed` fall through to `RETURN FALSE`. `confirmManualPayment` stamps the request `confirmed` **before** promoting, then throws "Subscription switch no longer matches the current subscription"; the plain-upsert fallback and tenant/revenue-split writes are gated on `!switch_id`, and `switch_id` never clears.

**Impact:** a school that actually wired the money can never be activated — every retry throws. The old unconditional guard plus unconditional upsert could always credit a late payment.

### 🔴 5. `isFreshActivation` bypasses the identity gate this PR adds

`lib/billing/platform-webhook-dispatch.ts:364`

`isFreshActivation` (`stored.status !== 'active'` + metadata `plan_id`) skips the non-current-identity gate. Tenant on Stripe `sub_A` in dunning (`past_due`): a delayed, never-processed `subscription.activated` for an old LS subscription (metadata carries this tenant's `plan_id`, no `billing_switch_id`, no `periodEnd` — so `isStalePeriod` can't drop it) passes as fresh, and the plain upsert rewrites `payment_provider`/`provider_subscription_id`/`status`/period wholesale.

**Impact:** `sub_A`'s recovery/cancellation events now fail `currentIdentity` and are ignored, while the stale identity's later `subscription.expired` passes the identity check and downgrades a paying tenant to free. Scoping the exception to terminal stored statuses (`canceled`/`expired`) instead of `!== 'active'` closes the hole.

### 🟠 6. An abandoned hosted checkout blocks all payment-method changes for 7 days

`app/api/billing/checkout/route.ts:263`

`replacementExpiresAt` is only computed for `provider === 'solana'`; Stripe and LS return `CheckoutSession` without `expiresAt`, so `attachSwitchCheckoutReference` writes nothing and the row keeps `now() + 7d`. The checkout `cancelUrl` just navigates back, `failPlatformSubscriptionSwitch` is only wired to server-side errors, and the cron's abandon sweep requires `expires_at < now`.

**Impact:** a school that mis-clicks "pay with Lemon Squeezy" and closes the tab cannot start ANY switch — same provider, different provider, or a manual request — for up to 7 days. If its subscription lapses in that window it downgrades to free with no way to re-pay.

### 🟠 7. `rejectManualPayment` never fails the linked switch row

`app/actions/platform/plans.ts:279`

Confirm handles the switch (promote + reconcile); reject only updates `platform_payment_requests.status`. The orphaned `pending_activation` row holds the one-open-switch unique index, so when the school fixes its bank reference and re-submits — exactly what a rejection asks it to do — `requestManualPlanUpgrade` and every provider checkout throw `SwitchAlreadyPendingError` until the cron abandons the row at day 14. The cron's Phase 0 request-expiry sweep has the same omission (expires the request, not its switch).

### 🟠 8. `alreadyCanceled()` decides idempotency by regex on error message text — wrong in both directions

`lib/billing/platform-subscription-switch.ts:163`

- **False positive → double-billing:** LS returns HTTP 422 "Store **not found** for this API key" (credential error, subscription untouched) — `/not found/i` matches, the catch synthesizes `{ mode: 'immediate' }`, the switch completes with `last_error = null`, and the LS subscription keeps auto-renewing alongside the new rail: the exact double-charge this PR exists to prevent, with the ledger asserting cleanup succeeded.
- **False negative → infinite retry:** `StripeProvider` wraps errors as "Stripe cancelSubscription failed: No such subscription: sub_x" — no pattern matches, so a source deleted at Stripe retries on 240-min backoff forever, holding the tenant's one-open-switch slot (`cancellation_retry` is in the unique index).

Providers should classify their own typed error codes and return/throw a typed AlreadyCanceled signal instead.

### 🟡 9. Serial provider cancels at the top of the cron can starve the money-critical phases

`app/api/cron/expire-platform-subscriptions/route.ts:138`

Switch reconciliation serially awaits up to 100 blocking provider HTTP cancel calls before Phase 0 even starts, and the route declares no `maxDuration`. A degraded provider at 5s/call × a 100-row backlog = 500s: no reminder emails, no grace windows, no downgrades — every run until the backlog drains, and a failing provider is precisely what grows the backlog. Bounded concurrency (or a smaller per-run cap) and moving the block after the existing phases removes the coupling.

### 🟡 10. The activation write now has twin writers in TypeScript and PL/pgSQL

`app/actions/admin/billing.ts:403`

`subscriptionValues` (16 columns plus the #546/#603 invariant comments) is built unconditionally but dead on the switch path, which re-implements the same activation write as ten args to the promote RPC. The next activation-semantics fix (the way #546 §1 added the un-cancel columns) gets edited into the visible TypeScript object, ships green, and provider-switching schools silently miss it. Same drift pattern in `downgrade-tenant.ts`, whose docstring ("Shared by the expiry cron and the Stripe platform webhook … so both agree") is now false — cron uses the TS writer, webhook the new SQL one.

---

guillermoscript (commented):
## ✅ Re-review of `343ec024` — all 10 findings verified fixed. Approve.

(Submitted as a comment because the PR author can't formally approve their own PR.)

Verified against the checked-out branch: **866/866 unit tests pass, `tsc --noEmit` clean.**

| # | Finding | Verification |
|---|---|---|
| 1 | Same-rail crypto renewal dropped by identity gate | ✅ `isSameRailSelfManagedRenewal` (same provider + `selfManagedPeriod`) rotates the synthetic sub id and extends the period; cross-provider events still excluded |
| 2 | `cancellation_scheduled` dead-end holds the one-open-switch lock | ✅ Removed from `platform_subscription_switches_one_open_per_tenant`'s WHERE list — a scheduled LS cancel no longer blocks future switches |
| 3 | Binance ~1h expiry → late webhook 500s forever | ✅ `promote_platform_subscription_switch` now revives `abandoned` switches when the source snapshot still exactly matches, and releases any newer unpaid pending intent so revival can't violate the unique index |
| 4 | Abandoned manual switch → `confirmManualPayment` throws forever | ✅ Same revival path covers late bank wires; plus new `rejected` guard on confirm |
| 5 | `isFreshActivation` bypasses the identity gate | ✅ Now requires no stored external identity, or same provider + terminal stored status (`canceled`/`incomplete_expired`) — a past_due tenant can no longer be overwritten by a stale foreign activation |
| 6 | Abandoned hosted checkout blocks all switches for 7 days | ✅ Default `expires_at` cut to 1 hour (residual below) |
| 7 | Reject/expiry orphans the linked switch | ✅ `rejectManualPayment` fails the switch; cron Phase 0 abandons switches of expired requests (status-gated so a concurrent confirm wins) |
| 8 | `alreadyCanceled()` message-regex idempotency | ✅ Regex deleted; each adapter classifies its own structured response — Stripe `resource_missing`/404, LS bare 404 (422 stays retryable), PayPal typed `PayPalApiError` 404 |
| 9 | Serial provider cancels starve money-critical cron phases | ✅ Cleanup moved to Phase 5 after expiry/reminders/grace/downgrades, batch capped at 10 |
| 10 | Twin activation writers (TS + PL/pgSQL) | ✅ As scoped in the design doc — `subscriptionValues` moved inside the non-switch branch (dead copy gone), `downgrade-tenant.ts` docstring now describes the real ownership split |

Bonus: the promote RPC now locks `platform_subscriptions` **before** reading the switch row — consistent lock order with `downgrade_platform_subscription_if_current`, eliminating the revival/new-checkout deadlock window.

### Non-blocking residuals (fine as follow-ups)

1. **Lockout is cron-cadence bound, not 1h.** `beginPlatformSubscriptionSwitch` still 23505s on an expired-but-unswept `pending_activation` row, and the abandon sweep only runs when the daily cron fires — a mis-clicked checkout can still block up to ~24h. Self-healing on conflict (abandon the expired row and retry the insert) would close it fully.
2. The abandon sweep's `.limit(100)` on an UPDATE bounds the returned representation, not the rows written — `switchesAbandoned` under-reports on a large backlog (observability only).

PR #628, opened 2026-08-08
fix(billing): atomically claim webhook events (#625)

### LINKED ISSUE #625: Atomically claim webhook events before dispatching billing side effects

guillermoscript:
## What

Atomically leases each student or platform billing webhook event before dispatch, fences completion/failure by a per-attempt token, and returns deterministic accepted, already-processing, or already-completed outcomes. It also records refund and self-managed-period effects transactionally so interleaved replay cannot duplicate accounting or paid access.

Closes #625

## Why

The previous check-then-insert flow allowed concurrent deliveries to reuse the same unprocessed ledger row and execute billing side effects more than once. A unique event key prevented duplicate rows, but not duplicate dispatch, while a completion write lost after successful dispatch could replay read-modify-write effects such as partial refunds and self-managed subscription extensions.

## How to QA

1. On a fresh local database, run `npm run db:reset` and export the local Supabase URL, anon key, and service-role key.
2. Run `npx playwright test tests/playwright/webhook-event-claims.spec.ts --project=desktop-chromium --workers=1 --retries=0`. Verify concurrent claims produce exactly one owner, stale tokens are fenced, explicit failure and lease expiry can be reclaimed, namespaces remain independent, and anonymous execution is denied.
3. Run `docker exec -i supabase_db_lms-front psql -v ON_ERROR_STOP=1 -U postgres -d postgres < tests/sql/issue-625-webhook-business-idempotency.sql`. Verify the transaction completes with `DO` and rolls back: A,B,A refund replay applies two slices, and A,B,A self-managed-period replay extends only twice.
4. Run `npm run test:unit`, `npm run typecheck`, and `npm run build`.
5. Deliver the same signed webhook twice while the first request holds its lease. Verify the second response is HTTP 409 with `eventStatus: "already_processing"` and `Retry-After: 30`; after completion, verify replay returns HTTP 200 with `eventStatus: "already_completed"` and performs no business writes.

## Screenshots / GIF

Not applicable. This change affects backend webhook processing and database behavior only.

## Checklist

- [x] `npm run build` passes
- [x] Every new tenant-scoped query filters by `tenant_id` (or the table genuinely has no such column)
- [ ] Tested with every relevant role (student / teacher / admin) — not applicable; these endpoints use provider signatures and the service role
- [ ] New UI strings added to both `messages/en.json` and `messages/es.json` — not applicable; no UI strings changed
- [ ] Migration (if any) applies cleanly on `npm run db:reset`, has RLS policies, and `lib/database.types.ts` was regenerated — migration applied cleanly to local PostgreSQL and has RLS/restricted RPC grants plus updated generated types; full reset was blocked by unrelated local migration-history drift (`20260808154118` exists in the local database but not this branch)

## Verification performed

- `npm run test:unit` — 65 files, 851 tests passed.
- `npm run typecheck` — passed.
- Targeted ESLint — no errors; one pre-existing `_cols` unused warning remains in `tests/unit/webhook-dispatch.test.ts`.
- `npm run build` — passed.
- `tests/playwright/webhook-event-claims.spec.ts` — 3/3 passed against local PostgreSQL.
- `tests/sql/issue-625-webhook-business-idempotency.sql` — passed with all mutations rolled back.
- `supabase db lint --local --level warning` — no findings in the new functions; existing unrelated schema findings remain in legacy functions.

### DIFF
--- a/app/api/payments/webhook/[provider]/route.ts
+++ b/app/api/payments/webhook/[provider]/route.ts
@@ -20,16 +20,15 @@ import {
   claimWebhookEvent,
   completeWebhookEvent,
   failWebhookEvent,
+  UNIFIED_STUDENT_WEBHOOK_PROVIDERS,
 } from '@/lib/payments/webhook-event-claim'
 
 export const runtime = 'nodejs'
 
-// Providers exposed on this endpoint. getPaymentProvider() still gates on
-// configured credentials; providers without verify/normalize return 501.
-// `manual` and `solana` are intentionally excluded: neither has a signed
-// webhook (Solana confirms on-chain via /api/payments/solana/verify), so
-// exposing a route for them would be an unauthenticated mutation surface.
-const SUPPORTED: PaymentProvider[] = ['stripe', 'paypal', 'lemonsqueezy', 'binance']
+// Providers exposed on this endpoint (shared with the redelivery cron; see the
+// definition for why manual/solana are excluded). getPaymentProvider() still
+// gates on configured credentials; providers without verify/normalize get 501.
+const SUPPORTED: PaymentProvider[] = UNIFIED_STUDENT_WEBHOOK_PROVIDERS
 
 /**
  * Provider-specific ACK body. Binance Pay treats anything other than
--- a/app/api/payments/webhook/[provider]/route.ts
+++ b/app/api/payments/webhook/[provider]/route.ts
@@ -132,7 +132,14 @@ export async function POST(
   }
   if (claim.status === 'processing') {
     return NextResponse.json(
-      ackBody(provider, { processing: true, eventStatus: 'already_processing' }),
+      provider === 'binance'
+        ? {
+            returnCode: 'FAIL',
+            returnMessage: 'Event is already processing',
+            processing: true,
+            eventStatus: 'already_processing',
+          }
+        : ackBody(provider, { processing: true, eventStatus: 'already_processing' }),
       { status: 409, headers: { 'Retry-After': '30' } },
     )
   }
@@ -155,6 +162,11 @@ export async function POST(
     await completeWebhookEvent(admin, claim)
   } catch (err) {
     console.error(`[webhook/${provider}] failed to complete event ${providerEventId}:`, err)
+    try {
+      await failWebhookEvent(admin, claim, err)
+    } catch (releaseErr) {
+      console.error(`[webhook/${provider}] failed to release claim after completion error:`, releaseErr)
+    }
     return NextResponse.json({ error: 'Event completion failed' }, { status: 500 })
   }
 
--- a/app/api/payments/webhook/[provider]/route.ts
+++ b/app/api/payments/webhook/[provider]/route.ts
@@ -16,6 +16,11 @@ import { createClient } from '@supabase/supabase-js'
 import { getPaymentProvider } from '@/lib/payments'
 import type { PaymentProvider } from '@/lib/payments/types'
 import { dispatchBillingEvent } from '@/lib/payments/webhook-dispatch'
+import {
+  claimWebhookEvent,
+  completeWebhookEvent,
+  failWebhookEvent,
+} from '@/lib/payments/webhook-event-claim'
 
 export const runtime = 'nodejs'
 
@@ -105,40 +110,31 @@ export async function POST(
 
   const admin = getSupabaseAdmin()
 
-  // 3. Idempotency: skip if this event was already processed.
-  const { data: existing } = await admin
-    .from('webhook_events')
-    .select('id, processed_at')
-    .eq('provider', provider)
-    .eq('provider_event_id', providerEventId)
-    .maybeSingle()
-
-  if (existing?.processed_at) {
-    return NextResponse.json(ackBody(provider, { duplicate: true }))
+  // 3. Atomically insert-or-claim before dispatch. Both student and platform
+  // webhook routes use this exact lease contract.
+  let claim
+  try {
+    claim = await claimWebhookEvent(admin, {
+      provider,
+      providerEventId,
+      eventType: event.type,
+      payload: event.raw as Record<string, unknown>,
+    })
+  } catch (err) {
+    console.error(`[webhook/${provider}] event claim failed:`, err)
+    return NextResponse.json({ error: 'Event claim failed' }, { status: 500 })
   }
 
-  let rowId = existing?.id as string | undefined
-  if (!rowId) {
-    const { data: inserted, error: insertErr } = await admin
-      .from('webhook_events')
-      .insert({
-        provider,
-        provider_event_id: providerEventId,
-        event_type: event.type,
-        payload: event.raw as Record<string, unknown>,
-      })
-      .select('id')
-      .single()
-
-    if (insertErr) {
-      // 23505 = unique violation: a concurrent delivery beat us to it.
-      if ((insertErr as { code?: string }).code === '23505') {
-        return NextResponse.json(ackBody(provider, { duplicate: true }))
-      }
-      console.error(`[webhook/${provider}] failed to persist event:`, insertErr)
-      return NextResponse.json({ error: 'Failed to persist event' }, { status: 500 })
-    }
-    rowId = inserted.id
+  if (claim.status === 'completed') {
+    return NextResponse.json(
+      ackBody(provider, { duplicate: true, eventStatus: 'already_completed' }),
+    )
+  }
+  if (claim.status === 'processing') {
+    return NextResponse.json(
+      ackBody(provider, { processing: true, eventStatus: 'already_processing' }),
+      { status: 409, headers: { 'Retry-After': '30' } },
+    )
   }
 
   // 4. Dispatch. On failure, record the error and 500 so the provider retries.
@@ -146,23 +142,21 @@ export async function POST(
     await dispatchBillingEvent(event, { provider, admin })
   } catch (err) {
     console.error(`[webhook/${provider}] dispatch failed:`, err)
-    await admin
-      .from('webhook_events')
-      .update({ error: err instanceof Error ? err.message : String(err) })
-      .eq('id', rowId)
+    try {
+      await failWebhookEvent(admin, claim, err)
+    } catch (releaseErr) {
+      console.error(`[webhook/${provider}] failed to release claim:`, releaseErr)
+    }
     return NextResponse.json({ error: 'Dispatch failed' }, { status: 500 })
   }
 
-  // 5. Mark processed. If this write fails the event will be redelivered and
-  //    re-dispatched; the dispatcher is idempotent (status writes converge,
-  //    period end comes from the event, not now()), so log rather than fail.
-  const { error: markErr } = await admin
-    .from('webhook_events')
-    .update({ processed_at: new Date().toISOString() })
-    .eq('id', rowId)
-  if (markErr) {
-    console.error(`[webhook/${provider}] failed to mark event ${providerEventId} processed:`, markErr)
+  // 5. Fence completion by token and fail closed if it is not durable.
+  try {
+    await completeWebhookEvent(admin, claim)
+  } catch (err) {
+    console.error(`[webhook/${provider}] failed to complete event ${providerEventId}:`, err)
+    return NextResponse.json({ error: 'Event completion failed' }, { status: 500 })
   }
 
-  return NextResponse.json(ackBody(provider))
+  return NextResponse.json(ackBody(provider, { eventStatus: 'accepted' }))
 }
--- a/lib/billing/platform-webhook-dispatch.ts
+++ b/lib/billing/platform-webhook-dispatch.ts
@@ -315,22 +315,33 @@ export async function dispatchPlatformBillingEvent(
   const periodEnd = event.periodEnd?.toISOString()
 
   if (event.type === 'subscription.past_due') {
-    await unwrap(
+    // The transition test lives in the WHERE clause, not in a read of `stored`:
+    // a single failed charge produces more than one past_due event (the failed
+    // invoice and the subscription's own status change), and two of them
+    // dispatching concurrently would BOTH pass a read-then-check against the
+    // same 'active' snapshot. Only the update that actually flips the row wins
+    // the right to notify.
+    const transitioned = (await unwrap(
       'platform_subscriptions past_due',
-      admin.from('platform_subscriptions').update({ status, updated_at: now }).eq('tenant_id', tenantId),
-    )
+      admin
+        .from('platform_subscriptions')
+        .update({ status, updated_at: now })
+        .eq('tenant_id', tenantId)
+        .neq('status', 'past_due')
+        .select('tenant_id'),
+    )) as { tenant_id: string }[] | null
     await unwrap(
       'tenants past_due',
       admin.from('tenants').update({ billing_status: status, updated_at: now }).eq('id', tenantId),
     )
 
-    // Only on the TRANSITION into dunning. A single failed charge produces more
-    // than one past_due event (the failed invoice and the subscription's own
-    // status change), and a redelivery produces another — mailing on each would
-    // send a school three copies of the same bad news. Best-effort besides: a
-    // mail failure must not undo the writes above by 500-ing, which would have
-    // the provider redeliver an event we had already applied.
-    if (stored?.status !== 'past_due') {
+    // Only on the TRANSITION into dunning — mailing on each event would send a
+    // school three copies of the same bad news. `!stored` keeps the pre-#625
+    // behavior of still warning a tenant that has no subscription row at all.
+    // Best-effort besides: a mail failure must not undo the writes above by
+    // 500-ing, which would have the provider redeliver an event we had already
+    // applied.
+    if ((transitioned?.length ?? 0) > 0 || !stored) {
       if (!event.providerEventId) {
         throw new Error(`platform dispatch ${event.type}: provider event id is required for notification dedupe`)
       }
--- a/lib/billing/platform-webhook-dispatch.ts
+++ b/lib/billing/platform-webhook-dispatch.ts
@@ -362,18 +362,6 @@ export async function dispatchPlatformBillingEvent(
     : undefined
   const interval = event.interval ?? mapInterval(event.metadata?.interval)
 
-  // Providers that bill on a schedule report the period they just charged for.
-  // The ones WE own report nothing, and a subscription with a NULL
-  // current_period_end never lapses, never reminds and shows no next-payment
-  // date — it is the whole difference between a paid month and a free one.
-  const derived =
-    selfManaged && !event.periodEnd && status === 'active'
-      ? selfManagedPeriod(stored?.current_period_end, interval, new Date(now))
-      : null
-
-  const effectiveStart = periodStart ?? derived?.start.toISOString()
-  const effectiveEnd = periodEnd ?? derived?.end.toISOString()
-
   // The plan the ROW must carry, which is a different question from `planId`
   // above. That one answers "is this event moving the school to a new plan?" and
   // is deliberately undefined on every event after the first; this one answers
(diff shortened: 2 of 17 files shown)


lib/payments/webhook-dispatch.ts
```diff
@@ -285,12 +285,21 @@ export async function dispatchBillingEvent(
...
 
-      // Only act on a completed purchase of one of the two kinds.
-      if (!tx || tx.status !== 'successful') break
+      // Durable provider-event refunds can revisit a fully refunded row to
+      // repair downstream entitlement state. Legacy callers without an event
+      // id keep the historical successful-only guard.
+      if (
+        !tx ||
+        (event.providerEventId
+          ? !['successful', 'refunded'].includes(tx.status)
```
guillermoscript: **[CONFIRMED] Refund branch re-revokes an entitlement the student re-purchased**

The refund branch acts on already-`'refunded'` transactions and ignores `apply_webhook_refund`'s `applied` flag, so a second refund-family event (or forced redelivery) re-runs the full-refund entitlement revoke and can strip access a student re-purchased.

**Failure scenario:** Full refund on tx 42 flips it to `'refunded'` and frees the partial unique index; the student re-buys and `enroll_user` reactivates the SAME entitlement row (`ON CONFLICT DO UPDATE`). A later distinct refund event for tx 42 — Lemon Squeezy mints a new providerEventId per delivery (`order_refunded:<orderId>:<updatedAt>`, so `webhook_business_effects` can't dedupe it), PayPal sends one `PAYMENT.CAPTURE.REFUNDED` per partial refund — reaches `apply_webhook_refund`, which returns `is_full_refund = (status='refunded') = true` even when `inserted_effect` is false; the TS never reads `applied` and revokes at line 358, killing the paying student's fresh entitlement.

supabase/migrations/20260808154215_atomic_webhook_event_claim.sql
```diff
@@ -0,0 +1,389 @@
...
+  if _new_period_end is null then return; end if;
+
+  update public.subscriptions
+  set end_date = _new_period_end,
+      current_period_end = _new_period_end,
+      subscription_status = 'active',
+      ended_at = null
+  where provider_subscription_id = _provider_subscription_id
+    and payment_provider = _provider
+    and (current_period_end is null or current_period_end <= _new_period_end)
```
guillermoscript: **Monotonic guard silently skips the whole activation/reactivation update**

`extend_subscription_period`'s new monotonic guard `current_period_end <= _new_period_end` silently skips the WHOLE update — including `subscription_status='active'`/`ended_at=NULL` and the entitlement reactivation — whenever the stored end is later than the event's, breaking the documented align-down-to-`renews_at` activation contract.

**Failure scenario:** `handle_new_subscription` pre-writes `current_period_end = start + plans.duration_in_days` (e.g. 31 days); the provider's real `renews_at` is earlier (30-day cycle, or February). `subscription.activated` / `subscription.renewed` calls `extend_subscription_period` with the earlier authoritative date → WHERE matches nothing → RETURNING NULL → status reset and entitlements UPDATE never run, no error surfaced, route ACKs, provider never retries. Billing page shows the wrong next-payment date, and a `past_due` subscription recovering via that event stays `past_due`. The old body (20260615140000) updated unconditionally; this semantic change to a shared student-subscription RPC ships inside a migration named for webhook claims.

lib/billing/platform-webhook-dispatch.ts
```diff
@@ -403,6 +391,37 @@ export async function dispatchPlatformBillingEvent(
...
+      effectiveStart = rows[0]?.period_start ?? undefined
+      effectiveEnd = rows[0]?.period_end ?? undefined
+    } else {
+      const derived = selfManagedPeriod(stored?.current_period_end, interval, new Date(now))
+      effectiveStart = derived.start.toISOString()
+      effectiveEnd = derived.end.toISOString()
+    }
+  }
+
   const subscriptionPatch: Record<string, unknown> = {
```
guillermoscript, line 563: **[CONFIRMED] Replay rewrites plan/cancel flags from stale metadata**

The effects ledger dedupes only the PERIOD on self-managed replay: a redelivered event still rewrites `plan_id`/`tenants.plan` from its stale metadata, re-applies the revenue split, and clears `cancel_at_period_end`/`canceled_at`/`grace_period_end` — the 'remains correct for A,B,A replay' comment holds for period columns only.

**Failure scenario:** Binance Starter payment dispatches fully but `completeWebhookEvent` 500s; lease expires; Binance redelivers and `claim_webhook_event` re-grants (`processed_at` NULL). Meanwhile the school upgraded to Pro or scheduled a cancel. Re-dispatch: `apply_self_managed_platform_period` returns `applied=false` (period safe) but the code never reads `applied` — `trustMetadataPlan` is always true for selfManaged, so the upsert writes `plan_id=Starter`, `tenants.plan='starter'`, `applyRevenueSplit(Starter)` re-applies the Starter fee, and `derivedPeriod` forces `cancel_at_period_end=false`. A Pro-paying school is gated and billed as Starter, or a scheduled cancel silently vanishes. The crypto-rails A,B,A unit test reuses one plan for A and B, so this is invisible to it.

app/api/payments/webhook/[provider]/route.ts
```diff
@@ -105,64 +110,53 @@ export async function POST(
...
-      console.error(`[webhook/${provider}] failed to persist event:`, insertErr)
-      return NextResponse.json({ error: 'Failed to persist event' }, { status: 500 })
-    }
-    rowId = inserted.id
+  if (claim.status === 'completed') {
+    return NextResponse.json(
+      ackBody(provider, { duplicate: true, eventStatus: 'already_completed' }),
+    )
+  }
+  if (claim.status === 'processing') {
```
Comment (guillermoscript) [inline app/api/payments/webhook/[provider]/route.ts:132]: **Dead worker's 300s lease can exhaust a provider's bounded retries**

A crashed worker (OOM/serverless kill — `failWebhookEvent` never runs) holds the lease for the full 300s, every provider retry in that window gets HTTP 409, and there is no sweeper for leased-but-dead rows — a provider with a bounded, front-loaded retry schedule can exhaust its retries and the event is never processed by anyone.

**Failure scenario:** Worker is killed mid-dispatch of a Binance PAY_SUCCESS. `processing_started_at` stays set; retries inside the 5-minute lease all 409 (non-2xx responses count toward Stripe's endpoint auto-disable and Binance's failing-webhook flag; the Binance 409 body even says `returnCode: SUCCESS` while the status says failure). If the provider's remaining redeliveries land inside the lease window, no delivery ever finds the lease expired, no cron reclaims stalled `webhook_events` rows, and the student paid but is never enrolled — where the old code reused the unfinished row on the very next delivery.

supabase/migrations/20260808154215_atomic_webhook_event_claim.sql
```diff
@@ -0,0 +1,389 @@
...
+  set processing_token = _claim_token,
+      processing_started_at = claim_time,
+      attempt_count = event.attempt_count + 1,
+      event_type = coalesce(_event_type, event.event_type),
+      payload = coalesce(_payload, event.payload)
+  where event.provider = _provider
+    and event.provider_event_id = _provider_event_id
+    and event.processed_at is null
+    and (
+      event.processing_started_at is null
```
guillermoscript, line 95: **Lease expiry judged by the challenger's lease — active claims are stealable**

Lease expiry in `claim_webhook_event` is judged against the CHALLENGER's `_lease_seconds`, not the lease the current owner claimed under, so any caller passing a short lease can steal an active claim and double-dispatch.

**Failure scenario:** The reclaim UPDATE checks `processing_started_at <= claim_time - make_interval(secs => bounded_lease_seconds)` using the incoming parameter. A claim granted with the standard 300s lease is stealable 1 second later by any call with `_lease_seconds=1` (the Playwright spec itself does `claim(token, 1)`). Latent today because both routes hardcode 300, but the first future caller that tunes `leaseSeconds` down starts re-dispatching events whose owners are still mid-flight. Store the lease duration on the row at claim time instead.

lib/payments/webhook-dispatch.ts
```diff
@@ -285,12 +285,21 @@ export async function dispatchBillingEvent(
...
+        )
         .eq('transaction_id', txnId)
         .maybeSingle()
 
-      // Only act on a completed purchase of one of the two kinds.
-      if (!tx || tx.status !== 'successful') break
+      // Durable provider-event refunds can revisit a fully refunded row to
+      // repair downstream entitlement state. Legacy callers without an event
+      // id keep the historical successful-only guard.
+      if (
```
guillermoscript: **Out-of-order refund on a pending transaction is silently marked completed**

A refund event arriving while its transaction is still `'pending'` (out-of-order delivery before the payment.succeeded flip) breaks out silently, and the route then marks the webhook event completed — the refund is never applied and every redelivery is ACKed as `already_completed`.

**Failure scenario:** Lemon Squeezy `order_refunded` lands before the `order_created`/`subscription_activated` flip, so `tx.status='pending'`: the guard breaks, `dispatchBillingEvent` returns normally, `completeWebhookEvent` latches `claim_status` to `'completed'`. `refunded_amount` is never recorded and `getPayoutsOwed` keeps paying the school for money the platform returned. Same silent-latch via `if (!result) break` at line 331 when `apply_webhook_refund` returns an empty set. For a PR whose purpose is durable refund accounting, this path should fail/release the lease rather than complete.

lib/billing/platform-webhook-dispatch.ts
```diff
@@ -403,6 +391,37 @@ export async function dispatchPlatformBillingEvent(
...
+          _tenant_id: tenantId,
+          _plan_id: rowPlanId,
+          _interval: interval ?? 'monthly',
+          _provider_subscription_id: event.providerSubscriptionId ?? null,
+          _provider_customer_id: event.providerCustomerId ?? null,
+        }),
+      )) as { period_start: string | null; period_end: string | null }[]
+      effectiveStart = rows[0]?.period_start ?? undefined
+      effectiveEnd = rows[0]?.period_end ?? undefined
+    } else {
```
guillermoscript: **Unreachable no-providerEventId fallbacks duplicate money-bearing logic**

Both no-providerEventId fallback branches are unreachable dead code (routes 422-reject events without `providerEventId`; Solana verify passes the signature; the legacy Stripe route handles refunds inline) and each keeps a second, drifting implementation of money-bearing logic instead of failing loudly.

**Failure scenario:** The TS selfManagedPeriod else-branch (here) and the legacy inline refund else-branch (`lib/payments/webhook-dispatch.ts:335`) preserve the exact racy read-modify-write semantics #625 exists to kill, with arithmetic that already disagrees with the SQL (`setUTCMonth` rolls Jan 31 → Mar 2/3; Postgres `interval '1 month'` clamps to Feb 28). A future internal caller that omits `providerEventId` silently resurrects the non-atomic double-apply bug rather than erroring; any future rule change must be made in two places per effect.

supabase/migrations/20260808154215_atomic_webhook_event_claim.sql
```diff
@@ -0,0 +1,389 @@
...
+    on conflict (tenant_id) do update set
+      plan_id = excluded.plan_id,
+      status = 'active',
+      payment_provider = excluded.payment_provider,
+      interval = excluded.interval,
+      provider_subscription_id = coalesce(excluded.provider_subscription_id, public.platform_subscriptions.provider_subscription_id),
+      provider_customer_id = coalesce(excluded.provider_customer_id, public.platform_subscriptions.provider_customer_id),
+      current_period_start = excluded.current_period_start,
+      current_period_end = excluded.current_period_end,
+      updated_at = excluded.updated_at
```
guillermoscript, line 375: **NULL-period return path lets the caller upsert a never-lapsing subscription**

`apply_self_managed_platform_period` returns NULL periods when the effect is already recorded but the `platform_subscriptions` row no longer exists, and the caller then upserts a fresh ACTIVE subscription row with no `current_period_end` — a subscription that never lapses.

**Failure scenario:** Admin resets a school's billing by deleting its `platform_subscriptions` row; `webhook_business_effects` rows survive. A late redelivery of an old payment event reaches dispatch (lease-expiry path): `inserted_effect=false`, `subscription_row` NULL → RPC returns NULL `period_start`/`period_end` → `effectiveStart`/`effectiveEnd` become undefined → `subscriptionPatch` upserts `status='active'`, `plan_id` set, `current_period_end` NULL. The deleted code's comment warned exactly this: a subscription with a NULL `current_period_end` never lapses, never reminds — the expire cron never downgrades it and the school sits on a paid plan free, indefinitely.

guillermoscript (commented):
## Code review — 10 findings (ranked by severity)

Automated review (adversarially verified; one candidate — the interval yearly→monthly overwrite — was refuted as unreachable and dropped). 8 findings are inline; 2 fall outside the diff hunks and are listed here:

---

### `lib/billing/platform-webhook-dispatch.ts:456` — Unguarded upsert re-races the period the RPC just serialized

The period stacking serialized inside `apply_self_managed_platform_period` is immediately re-written by an unguarded `platform_subscriptions`/`tenants` upsert in a separate transaction, so two concurrent different payments for the same tenant can still rewind `current_period_end` — the exact loss the RPC was added to prevent.

**Failure scenario:** Events A and B (two renewals) dispatch concurrently. The RPC serializes correctly under `pg_advisory_xact_lock`: A → E1, B stacks → E2. The lock is xact-scoped and releases when each RPC's PostgREST transaction commits; each worker then upserts its own returned `effectiveEnd` (lines 456–461) and `tenants.billing_period_end` (line 484) with no monotonic guard. If A's upserts land after B's, the row rewinds E2 → E1: the school paid two periods, got one, and the expire cron acts on the rewound date. `isStalePeriod` can't help (self-managed events carry no `event.periodEnd`). The RPC and `subscriptionPatch` each own overlapping columns with different rules — make the RPC the single writer or re-check monotonicity in the upsert.

---

### `lib/billing/platform-webhook-dispatch.ts:333` — Side effects run before the completion fence; duplicate dunning emails

A slow worker whose lease expires re-executes every effect not covered by `webhook_business_effects` concurrently with its reclaimer — concretely, duplicate payment-failed dunning emails to every school admin.

**Failure scenario:** Worker 1 claims `subscription.past_due`, reads `stored.status='active'`, stalls >300s on SMTP. Worker 2 reclaims, also reads `'active'` (worker 1 hasn't written), emails all admins, writes `past_due`, completes. Worker 1 wakes past its own `stored?.status !== 'past_due'` check and emails everyone again; only then does its `completeWebhookEvent` get fenced. Status writes converge; emails do not. Completion failure at route line 156 also leaves the lease held (`failWebhookEvent` is not called), guaranteeing a later full re-dispatch of all non-ledgered effects.

---

lib/payments/webhook-dispatch.ts
guillermoscript: ✅ Addressed in a1a754f2: `apply_webhook_refund` now early-returns `applied=false` for any non-`successful` row (an already-`refunded` transaction never reaches the revoke), and the entitlement revocation runs only when THIS event's ledger effect row was inserted AND the update itself flipped the row — dedupe, accounting and revoke share one transaction, so a redelivered or per-slice refund event can no longer strip a re-purchased entitlement.

supabase/migrations/20260808154215_atomic_webhook_event_claim.sql
guillermoscript: ✅ Addressed in a1a754f2: `extend_subscription_period` is replaced by `apply_webhook_subscription_period` with `_allow_period_realign` — activation passes `true` (align down to the provider's renews_at, per the documented contract), renewal passes `false` (monotonic via `greatest()`). The `subscription_status='active'`/`ended_at=NULL` reset and the entitlement reactivation now run unconditionally once the event's effect is claimed, so a shorter authoritative period can no longer silently skip a past_due recovery.

lib/billing/platform-webhook-dispatch.ts
guillermoscript: ✅ Addressed in a1a754f2: the plan write, `tenants` update, revenue split and cancellation reset all moved inside `apply_self_managed_platform_period`, gated on `inserted_effect` — a replayed event's stale metadata can no longer rewrite `plan_id`/`tenants.plan`, re-apply the split, or clear a scheduled cancel. The TS self-managed path returns before the generic upserts.

app/api/payments/webhook/[provider]/route.ts
Comment (guillermoscript) [reply app/api/payments/webhook/[provider]/route.ts:132]: ✅ Addressed in a1a754f2 + 5e82faf4: the Binance 409 body now says `returnCode: FAIL` (no more mixed signal) with a `Retry-After: 30` header, completion failure releases the lease, and 5e82faf4 adds `/api/cron/redeliver-webhook-events` (every 10 min) — the missing sweeper. It replays unprocessed rows whose lease expired (crashed worker) or that were released with an error, through the same normalize → claim → dispatch → complete pipeline, so a provider that exhausts its retries inside the lease window no longer strands a paid event. Poison events park at 8 attempts with `last_error` retained.

supabase/migrations/20260808154215_atomic_webhook_event_claim.sql
guillermoscript: ✅ Addressed in a1a754f2: the lease duration is stored on the row at claim time (`processing_lease_expires_at`) and challengers are judged against the OWNER's stored expiry, not their own `_lease_seconds` — a short-lease caller can no longer steal an active claim.

lib/payments/webhook-dispatch.ts
guillermoscript: ✅ Addressed in a1a754f2: a refund arriving while the transaction is still `pending` now throws (the RPC raises `40001`), and a missing transaction throws too — the lease is released, the route 500s, and the provider redelivers after the succeeded flip lands. No more silent `already_completed` latch over an unapplied refund.

lib/billing/platform-webhook-dispatch.ts
guillermoscript: ✅ Addressed in a1a754f2: both no-`providerEventId` fallback branches were deleted — the dispatchers now throw when the id is missing, so the money-bearing period/refund arithmetic has exactly one implementation (the SQL RPCs) and a future caller that omits the id fails loudly instead of resurrecting the racy read-modify-write.

supabase/migrations/20260808154215_atomic_webhook_event_claim.sql
guillermoscript: ✅ Addressed in a1a754f2: on a replay whose `platform_subscriptions` row no longer exists, the RPC raises `P0002` instead of returning NULL periods, and the TS side throws if the RPC returns no durable period — the caller no longer upserts anything on the self-managed path, so the never-lapsing NULL-`current_period_end` subscription can't be created.

guillermoscript:
Follow-up on the two review-body findings (the ones outside the diff hunks), resolved across a1a754f2 and 5e82faf4:

- **`platform-webhook-dispatch.ts:456` — unguarded upsert re-races the RPC:** fixed in a1a754f2 by making `apply_self_managed_platform_period` the single writer for the self-managed subscription row, `tenants.billing_period_end`, plan, cancellation reset and revenue split. The TS path now only upserts `tenant_billing_customers` and returns — no worker can rewind a newer serialized period.
- **`platform-webhook-dispatch.ts:333` — effects before the fence / duplicate dunning emails:** a1a754f2 fenced the email behind a `webhook_business_effects` claim (same-event reclaim can't re-mail) and made completion failure release the lease in both routes; 5e82faf4 closes the residual cross-event race by moving the transition test into the WHERE clause (`.neq('status','past_due').select()`) — of two concurrent different past_due events, only the update that actually flips the row sends the email.

5e82faf4 also adds the missing sweeper: `/api/cron/redeliver-webhook-events` replays dead-lease / released-with-error events through the same claim pipeline every 10 minutes (unit-tested in `tests/unit/redeliver-webhook-events.test.ts`). Full unit suite (864), typecheck and production build pass.

PR #633, opened 2026-08-21
feat(ai): let students attach images in every AI chat

guillermoscript:
## Summary
Students were attaching images to the AI chats and nothing happened: the lesson tutor showed the preview but `sendMessage({ text })` dropped `message.files`; the exercise coach, Aristotle and the teacher preview had no attach control at all. The routes already run `convertToModelMessages` on `gpt-4o-mini` (vision-capable), so the break was client-side plus text-only persistence.

- `components/ai/chat-attachments.tsx` — shared attach button, inline preview, image-part renderer, client-side downscale (≤1600px; `useChat` resends the whole history inline every turn)
- Every `PromptInput` accepts `image/*` only, ≤4 files, ≤5 MB, with error toasts
- `lib/ai/attachments.ts` — server sanitiser (body is user-controlled), upload to private `ai-chat-attachments` bucket, signed-URL re-hydration; earlier turns' signed URLs become a text note so an expired URL can't fail a model call
- Migration `20260821120000`: bucket + `attachments jsonb` on `exercise_messages` / `lessons_ai_task_messages`; image-only turns are now persisted
- Exercise page hydrated legacy `{content}` rows without `parts` (would crash `message.parts.map`) — fixed

## QA
Browser-verified on the exercise coach: image renders in the user bubble, model described it accurately, DB row + storage object written, reload re-hydrates via signed URL, follow-up turn works. `npm run typecheck`, `npm run build`, and 7 new unit tests (`tests/unit/ai-attachments.test.ts`) pass.

**Deploy note:** apply the migration to cloud (`supabase db push` or MCP) before/with the deploy — inserts send `attachments` and will 400 without the column.

https://claude.ai/code/session_01Q4Sh3N4uXFAug3dVHrQ4og


guillermoscript:
## Follow-up: fixes so the feature survives a real upload

Reviewed the branch end to end and pushed `79dc714a`. Four things stood between this and shipping.

### 1. ⚠️ Blocker — the migration is not on cloud (nothing in this push fixes it)

`supabase/migrations/20260821120000_ai_chat_attachments.sql` is applied locally but **not on the cloud project**: no `attachments` column on either table, no `ai-chat-attachments` bucket.

This is worse than "attachments don't work". Both student pages now select the column:

- `app/[locale]/dashboard/student/courses/[courseId]/exercises/[exerciseId]/page.tsx` → `exercise_messages(..., attachments)`
- `app/[locale]/dashboard/student/courses/[courseId]/lessons/[lessonId]/page.tsx` → `lessons_ai_task_messages(..., attachments)`

Deploying without the migration makes PostgREST 42703 those queries, so **the exercise page and the lesson page break entirely for every student** — not just the images. The migration must land before or with the deploy. I was blocked from applying it to cloud from here.

### 2. iPhone photos vanished without a trace

The picker accepted `image/*`; the routes only forward jpeg/png/gif/webp. A HEIC attached fine, rendered a chip, sent — and `sanitizeLastUserAttachments` dropped it server-side with no error surfaced anywhere. The student sees the model answer a question about an image it never received.

`prepareChatFiles` now transcodes any `image/*` the model can't read into JPEG via the canvas it was already using to downscale.

### 3. Ordinary camera photos were refused at pick time

`maxFileSize` was the 5 MB wire limit, checked *before* downscaling — but a phone photo is routinely 8–15 MB off the camera and shrinks to a few hundred KB at 1600px. The most common upload was rejected outright.

Split into two numbers: **25 MB at the picker** (pre-downscale) and **5 MB on the wire** (post-downscale, matching what the server enforces on decoded bytes). A resize ladder (1600 → 1200 → 900 → 700) runs until the encoded image fits, so a 48 MP panorama still gets through.

### 4. Silent drops and hardcoded English

`prepareChatFiles` returns `{ files, rejected }`; `useAiChatSubmit` names each rejected image in a toast and won't send a turn that has nothing left in it. All attachment copy moved to `components.aiAttachments` in en/es — `chatAttachmentInputProps` became `useChatAttachmentInputProps()` so it can reach next-intl. Dropped the duplicate `lessonAIChat.tooltips.attachImage`.

### Verification

| Check | Result |
|---|---|
| `npm run build` | ✅ pass |
| `npx tsc --noEmit` | ✅ clean |
| `eslint` (touched files) | ✅ clean |
| `npm run test:unit` | ✅ 870 pass; attachment specs 8 → 15 cases |
| Browser QA | ❌ **not done** |

**Browser QA is unverified.** The local Docker daemon wedged after the machine filled its disk mid-session, so I could not bring Supabase + dev up to drive a real upload. Nothing here is confirmed against a live round-trip.

Worth doing manually before merge, on `default.lvh.me` as `student@e2etest.com`:

1. Attach a JPG to an exercise chat → it renders in the bubble, the model responds to it, and `exercise_messages.attachments` holds a `{path, mediaType, filename}` row.
2. Reload the page → the image re-hydrates from a signed URL rather than the turn going text-only.
3. Attach a HEIC (or any iPhone photo) → it arrives as JPEG rather than disappearing.
4. Attach a PDF → a toast names it; the turn is not sent empty.
5. Same on the lesson tutor, and on Aristotle (whose history is intentionally not re-hydrated — image-only turns persist as `[image]`).

Issue #606, opened 2026-08-05
Student checkout accepts a school whose Stripe onboarding was abandoned


guillermoscript:
# Student checkout accepts a school whose Stripe onboarding was abandoned

Part of #600.

## Problem

`app/api/stripe/create-payment-intent/route.ts:73-88` gates student checkout on the *presence* of a Connect account id:

```ts
.from('tenants').select('stripe_account_id')
...
if (!tenant.stripe_account_id) {
  return NextResponse.json({ error: 'School has not connected their payment account...' }, { status: 400 })
}
```

But `stripe_account_id` is written the moment the Express account is **created**, before any KYC is done — `app/api/stripe/connect/route.ts:60-77` calls `stripe.accounts.create` then immediately persists the id, and only then hands the admin the hosted onboarding link. An admin who opens onboarding and closes the tab leaves a tenant that looks payment-ready and is not.

The readiness signal already exists and is already synced: `tenants.stripe_charges_enabled` / `stripe_payouts_enabled` / `stripe_details_submitted`, added in `20260717160000_add_stripe_connect_status_to_tenants.sql` and updated by the `account.updated` webhook (`app/api/stripe/webhook/route.ts:444-458`). `app/[locale]/dashboard/admin/settings/page.tsx:69-75` reads all three. **Checkout is the one place that doesn't.**

Failure mode: the student reaches the card form, submits, and Stripe rejects the PaymentIntent because `transfer_data.destination` (line 229) points at an account that cannot accept charges. The student sees a generic payment failure on the school's storefront and there is nothing in the product telling either party why.

## Wanted

- Gate on readiness, not existence: require `stripe_charges_enabled`, with a distinct error from the "never connected" case so the UI can say *"this school is still finishing payment setup"* rather than *"contact the admin"*.
- Express as an ability, not a Stripe special case: a `requiresConnectedAccount` capability on `ProviderCapabilities`, checked through one `isReadyToAcceptPayments(tenantId)` helper. Only Stripe Connect needs it today (`settlesToPlatformAccount: false` + a per-tenant account); `manual`, the platform-settled trio and the Solana rails each answer differently, and the branch should read off the capability rather than the provider name — the rule the table's own header states.
- Same gate on the product/plan **publish** path, so a school cannot publish a paid Stripe offering it cannot be paid for. `app/actions/admin/settings.ts` already exposes provider enablement; this is the missing consistency check next to it.
[...]

PR #643, opened 2026-08-29
feat(admin): rebuild Settings → Payments as per-provider connection rows

guillermoscript:
## What

Rebuilds admin **Settings → Payments** from a flat list of six toggle switches into per-provider connection rows, and closes three server-side gaps where a rail could be offered at checkout that could not actually take money.

## Why

The tab modelled six providers as six identical binary switches. They aren't — each is a connection with state, a credential surface, a settlement route and a fee.

**The reported bug.** Stripe was the only provider that got a state UI, and it was bolted on as a page-level alarm. `StripeConnectCard` never received `stripeEnabled`, so a school with Stripe switched **off** still saw "Stripe Not Connected" as the loudest element on the page — in the tenant's own brand colour, above a processor it had deliberately disabled.

**The inverse, which cost real money.** `getEnabledPaymentProviders()` pushed `stripe` and `solana`/`solana_subs` on the `tenant_settings` flag alone. Toggle Stripe on without finishing Connect onboarding, or Solana on with no receiving wallet, and those rails appeared in the product wizard and at student checkout, then failed at payment time. Only `binance_personal` had a readiness gate.

**A third gap found on the way.** `binance_personal` required only `api_key`. Confirming a transfer signs the SAPI query with the **secret** (`signSapiQuery`), so a half-saved credential pair could take a payment it could never confirm — while `getBinancePersonalStatus()`, which drives the settings screen, always demanded both. The screen and checkout disagreed.

## How

**Server** (`app/actions/admin/settings.ts`)
- `stripe` gated on real readiness via `evaluateConnectedAccountReadiness` from `lib/payments/tenant-payment-readiness.ts` (#606), so settings, checkout and the publish path share one definition of "ready". Reads the persisted columns — no Stripe API call on this hot path.
- `solana`/`solana_subs` gated on a `tenant_payment_wallets` row with a `wallet_address`.
- `binance_personal` now requires `api_key` **and** `api_secret`.
- The two wallet lookups collapse into one query instead of one per provider.

**UI**
- New `lib/payments/provider-presentation.ts` derives each row's fact chips from `PROVIDER_CAPABILITIES` rather than restating them by hand, so the screen cannot drift from what checkout will really do. It surfaces the fact the old screen hid entirely: PayPal, Lemon Squeezy and Binance Pay settle into a platform account and are paid out later, while Stripe, Solana, Binance personal and offline pay the school directly.
- Rows carry a status pill: **Ready / Action needed / Off / Not set up**. "Action needed" is reachable only from `enabled && !configured` — the state that silently loses sales. **A rail that is off never alarms.**
- Grouped **Cards & wallets / Crypto / Offline**.
- Offline gets a real row. It was always live in `getEnabledPaymentProviders()` yet absent from the list, with its instructions textarea and approval toggle stranded in two unrelated places.
- Solana wallet and Binance credentials move from trailing cards (each with its own save button) into sheets opened from their provider row. Three competing saves become one.
- Provider hints stop restating the provider's name and say what the student actually pays with.
- a11y: `aria-describedby` covers description + facts, rows are width-capped so the switch isn't ~1400px from its label, and the locked offline row states its status instead of rendering a dead switch.

`stripe-connect-card.tsx` is deleted — all three of its states now live in the Stripe row.

Types: the pre-commit lint gate forced replacing `Record<string, any>` with `SettingValue` / `SettingEntry` / `SettingsGroup`, and splitting the wallet accessors onto their own `WalletResponse`. That surfaced three latent mistypings (a wallet row was typed as a setting value).

## How to QA

On a fresh `npm run db:reset`, `npm run dev`, sign in as `owner@e2etest.com` / `password123` at `default.lvh.me:3000`, go to `/en/dashboard/admin/settings?tab=payment`.

1. **The reported bug is gone** — there is no page-level "Stripe Not Connected" banner. Stripe's state lives in the Stripe row.
2. Stripe shows **Action needed** (it is on by seed default and has no Connect account) with an inline "Connect Stripe" and a red one-liner. Toggle Stripe **off** → the pill goes muted "Not set up" and the red text disappears. That's the fix: off never alarms.
3. Toggle **Solana** on → pill flips "Not set up" → "Action needed", a red "Add your receiving details below" appears, and the "Also accept native SOL" sub-toggle expands inline.
4. Click **Add details** on Solana → a right-hand sheet opens with the wallet form (it saves itself).
5. Confirm the offline row exists under **Offline**, marked "Always available", holding both the instructions textarea and the approval toggle.
6. Confirm exactly **one** "Save changes" button in the form.
7. Repeat on `/es/...` — every string is translated.

## Screenshots / GIF

Verified live in a browser at 1440px; screenshots are in the session, happy to attach if you want them on the PR.

Status pills observed on seed data: Stripe `Action needed` · PayPal `Off` · Lemon Squeezy `Off` · Solana `Not set up` · Binance Pay `Off` · Binance Pay personal `Not set up` · Offline `Always available`. Zero page errors, zero failed requests, no missing-i18n keys.

## Checklist

- [x] `npm run typecheck` and `npm run test:unit` pass — 1000 passed (77 files), incl. 6 new cases pinning the readiness gates
- [x] `npm run build` passes — **confirmed green in CI** (`Production image builds` ✓). It failed locally only because my `node_modules` had `stripe@22.4.0` while the lockfile pins `22.6.0`; CI installs from the lockfile, so the pre-existing Stripe-pin type errors in `lib/stripe.ts` / `lib/payments/stripe-provider.ts` don't occur there.
- [x] Tested as admin (the only role that can reach this page; non-admins are redirected)
- [x] New UI strings added to both `messages/en.json` and `messages/es.json` (key parity verified both ways)
- [x] No migration

https://claude.ai/code/session_019s4zMecWkSRVSHD9dVTKE8


guillermoscript:
### Follow-up found while building this, deliberately out of scope

This PR gates `getEnabledPaymentProviders()` on real readiness, but there's a **fourth** instance of the same class of bug that it only partly mitigates. Filing it here so it isn't lost.

**The Solana wallet check never reaches the shared readiness module.**

`app/api/payments/checkout/route.ts:135` calls `isReadyToAcceptPayments(tenantId, 'solana')`. That's a **no-op** for Solana — `requiresConnectedAccount: false`, so it short-circuits to ready with zero queries. The route then resolves settlement (USDC/SOL at the live rate), snapshots the revenue split, **inserts a pending `transactions` row and returns a QR payload** — all before anything checks whether the school has a receiving wallet. The missing-wallet `400` only fires later, when the student's wallet app hits `app/api/payments/solana/tx/route.ts:85-92`.

No money moves (nothing settles until `/verify`), so this isn't a loss bug like the Stripe one. But every unconfigured-wallet Solana checkout leaves a **dangling `pending` transaction row** and burns a full round trip presenting a QR that cannot work.

The wallet check is currently hand-rolled in three separate places, none of them going through `tenant-payment-readiness.ts`:
- `app/api/payments/solana/tx/route.ts:85-92`
- `app/actions/admin/products.ts:166-176`
- `app/actions/admin/plans.ts:113-123`

**Suggested fix (not done here):** add a `requiresReceivingWallet` capability to `ProviderCapabilities` in `lib/payments/types.ts` (true for `solana`/`solana_subs`), and teach `tenant-payment-readiness.ts` to branch on it in parallel with `requiresConnectedAccount`, returning a new `no_receiving_wallet` reason. Checkout, publish and `getEnabledPaymentProviders` would then share one gate, and those three inline checks collapse into it.

Two wrinkles for whoever picks it up:
1. `lib/payments/payment-readiness-codes.ts` only has `not_connected` / `onboarding_incomplete`, and both `READINESS_MESSAGE` and `READINESS_ADMIN_MESSAGE` are Stripe-worded ("Connect your Stripe account", "card payments"). A wallet case needs provider-agnostic or per-provider copy.
2. A plain boolean capability won't cover `binance_personal`, which needs `wallet_address` **and** `api_key` **and** `api_secret` (this PR fixes the missing-secret half of that gate).

**What this PR does change here:** an unconfigured Solana rail no longer appears in the product/plan wizard at all, so reaching the dangling-row path now requires an *existing* Solana product whose wallet was cleared afterwards. Narrower, but still open.

PR #645, opened 2026-08-30
chore(deps): bump the AI SDK to 7.0.85

guillermoscript:
## What

Routine bump of the `ai` package (7.0.48 → 7.0.85) and the four `@ai-sdk/*` providers we use, plus the transitive `@ai-sdk/gateway` / `provider` / `provider-utils` packages they pull in.

Closes #

## Why

The manifest floors had drifted well behind what was actually installed locally. This pins them to the versions that are now in the lockfile, so a fresh clone installs what we tested rather than something older.

The lockfile diff is contained to the ai-sdk family — nothing else moves. In particular the exact `stripe` pin from #638 and its `apiVersion` literal are untouched.

## How to QA

No behaviour change is expected; this is a dependency-only PR. To confirm the AI surfaces still work end to end:

1. `npm ci && npm run db:reset`, then `npm run dev`.
2. Sign in as `student@e2etest.com` / `password123` on `default.lvh.me:3000`.
3. Open any lesson and use the AI chat — send a message with and without an image attachment (the flow from #633). Streaming should be unchanged.
4. Open an exercise and submit it for AI evaluation; confirm feedback still streams back.

## Screenshots / GIF

n/a — no user-visible change.

## Checklist

- [x] `npm run typecheck` and `npm run test:unit` pass (1000/1000)
- [x] `npm run build` passes
- [x] Every new tenant-scoped query filters by `tenant_id` (or the table genuinely has no such column) — no queries touched
- [x] Tested with every relevant role (student / teacher / admin) — n/a, no behaviour change
- [x] Loading and error states handled — n/a
- [x] Migration (if any) applies cleanly on `npm run db:reset` — no migration

https://claude.ai/code/session_01CZmStQ3YnZMQkr2sp2FeGr

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,10 +11,10 @@
         "packages/*"
       ],
       "dependencies": {
-        "@ai-sdk/google": "^4.0.31",
-        "@ai-sdk/openai": "^4.0.27",
-        "@ai-sdk/otel": "^1.0.48",
-        "@ai-sdk/react": "^4.0.51",
+        "@ai-sdk/google": "^4.0.48",
+        "@ai-sdk/openai": "^4.0.45",
+        "@ai-sdk/otel": "^1.0.71",
+        "@ai-sdk/react": "^4.0.74",
         "@base-ui/react": "^1.6.0",
         "@codesandbox/sandpack-react": "^2.20.0",
         "@dnd-kit/core": "^6.3.1",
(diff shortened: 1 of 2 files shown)


guillermoscript:
Superseded by #708, which carries higher versions of every package this PR bumps:

| package | this PR | on `master` after #708 |
|--|--|--|
| `ai` | 7.0.85 | 7.0.94 |
| `@ai-sdk/react` | 4.0.88 | 4.0.97 |
| `@ai-sdk/google` | 4.0.58 | 4.0.65 |
| `@ai-sdk/openai` | 4.0.52 | 4.0.62 |
| `@ai-sdk/otel` | 1.0.85 | 1.0.94 |

Nothing left to merge here.

PR #646, opened 2026-08-30
refactor(lesson): flip the flashcard with one face instead of two

guillermoscript:
## What

Rewrites `FlashcardSet`'s flip so only one face is ever in the DOM. Instead of stacking a front and a back and spinning the wrapper 180° behind `preserve-3d` + `backface-visibility: hidden`, the card rotates to `rotateY(90deg)` (edge-on, nothing legible), swaps its text there, and rotates back.

Closes #

## Why

**Read this before approving — the change is honest but its original premise is not verified.**

This change was sitting uncommitted in my working tree, written to fix a reported symptom: the two faces overlapping, or the card blanking out, when an ancestor flattens the 3D context. I could not reproduce that symptom:

- An isolated harness with the old markup under a flattening ancestor (`overflow-y: auto` + `filter`) renders correctly in **both Chromium 1234 and WebKit 26.5**.
- The lesson shell that actually renders this component — the `<article className="prose …">` in `lesson-content.tsx` — has no `overflow`, `filter` or `transform` of its own, so there is no flattening ancestor in the real page either.

So I am **not** claiming this fixes a rendering bug. What it does stand on:

- It is simpler — one face, one element, no dependency on the 3D context surviving whatever wraps the MDX.
- The timer is cleared on unmount and on card navigation, so a flip in flight cannot rewrite the card you just moved to, and a repeat click is ignored mid-flip.
- Reading `prefers-reduced-motion` moves to the `useSyncExternalStore` pattern `feedback-button.tsx` already uses. The old setState-in-effect trips `react-hooks/set-state-in-effect`, which the pre-commit hook enforces — the file could not be committed without this.

The trade-off: the animation is no longer a true 180° spin, it is a half-flip with a text swap at the midpoint. Visually close, not identical.

**Draft** until someone confirms the original symptom is real. If it was never real, the honest options are to close this or land it purely as the lint + timer cleanup.

## How to QA

The local Supabase stack was down (Docker daemon not running) when I wrote this, so I could not capture the component in the running app. No lesson in `supabase/seed.sql` uses a `<FlashcardSet>` block, so QA needs one authored first:

1. `npm run db:reset && npm run dev`.
2. As a teacher, add a `<FlashcardSet cards={[{front:"…",back:"…"},{front:"…",back:"…"}]} />` block to any lesson's MDX.
3. As `student@e2etest.com` / `password123` on `default.lvh.me:3000`, open the lesson.
4. Click the card — it should turn edge-on, swap text, and turn back. Click again to return.
5. Flip, then immediately hit the next/prev arrow: the new card must show its **front**, not the previous card's back.
6. Enable Reduce Motion in System Settings and reload: the swap should be instant, with no rotation.
7. Repeat in Safari — that is the browser the 3D-flattening concern was about.

## Screenshots / GIF

Not included on purpose. I have an isolated before/after harness, but since it does not reproduce the failure it would be misleading to present it as proof. Real in-app capture is pending a working local stack.

## Checklist

- [x] `npm run build` passes
- [x] Every new tenant-scoped query filters by `tenant_id` (or the table genuinely has no such column) — no queries touched
- [ ] Tested with every relevant role (student / teacher / admin) — **not done, local stack down**
- [x] Loading and error states handled — n/a, component renders `null` with no cards
- [x] New UI strings added to both `messages/en.json` and `messages/es.json` — none added (the two `aria-label`s are pre-existing hardcoded Spanish, untouched)
- [x] Migration (if any) applies cleanly on `npm run db:reset` — no migration

https://claude.ai/code/session_01CZmStQ3YnZMQkr2sp2FeGr

### DIFF
--- a/components/lesson/flashcard-set.tsx
+++ b/components/lesson/flashcard-set.tsx
@@ -1,6 +1,6 @@
 'use client'
 
-import { useState, useEffect } from 'react'
+import { useState, useEffect, useRef, useSyncExternalStore } from 'react'
 import { cn } from '@/lib/utils'
 import { IconChevronLeft, IconChevronRight } from '@tabler/icons-react'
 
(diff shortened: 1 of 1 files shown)


guillermoscript:
Closing. The premise didn't hold up.

The change was written to fix the two faces overlapping / the card blanking out when an ancestor flattens the 3D context. I couldn't reproduce that:

- the old markup under a flattening ancestor (`overflow-y: auto` + `filter`) renders correctly in both Chromium 1234 and WebKit 26.5
- the `<article className="prose …">` in `lesson-content.tsx` that actually renders this component has no `overflow`, `filter` or `transform` of its own — there's no flattening ancestor in the real page

Without a real symptom this is a behaviour change (true 180° spin → half-flip with a midpoint text swap) in exchange for nothing, so it isn't worth the review.

The one thing here that was genuinely blocking is unrelated to the flip: `components/lesson/flashcard-set.tsx:24` trips `react-hooks/set-state-in-effect`, so the pre-commit hook rejects the file on any edit. If that becomes annoying, the fix is the `useSyncExternalStore` pattern already in `components/shared/feedback-button.tsx:81-89` — worth a small standalone PR whenever someone next needs to touch this file.

PR #707, opened 2026-09-08
chore(deps): Next.js 16.2.12 → 16.3.4

guillermoscript:
## What

Bumps `next` and `eslint-config-next` from **16.2.12 → 16.3.4** (minor). Exact pins kept, matching the existing convention for these two.

`package-lock.json` shrinks by ~570 net lines — Next 16.3 dedupes several transitive deps.

## Why

Routine catch-up; 16.2.12 is two minors behind. No app code changes needed.

## QA

- `npm run typecheck` — clean
- `npm run build` — compiled successfully, full route table unchanged
- `npm run test:unit` — 85 files / 1077 tests passing

Not run here: Playwright E2E (needs a local Supabase + dev server).

https://claude.ai/code/session_01A6ghiHrLpvdQpcFzaTNi4V


guillermoscript:
## Blocked — 16.3.4 breaks the student dashboard in dev

Found while QAing the stacked sweep (#708). Putting this back to draft rather than merging it.

**Symptom.** Log in as a student, land on `/en/dashboard/student` via the post-login client-side navigation, and the page throws in the browser:

```
TypeError: this.value.push is not a function
```

React unwinds to `app/[locale]/error.tsx` and the user gets "Something went wrong" with a 500 shell. `getByTestId('student-dashboard')` never appears, so `gamification.spec.ts` times out on its `beforeEach`.

**Bisected to Next itself, not the sweep.** This branch carries only `next` + `eslint-config-next`, and it reproduces here. Reinstalling `origin/master`'s lockfile (next 16.2.12) on the same working tree makes the same test pass.

| build | result |
|--|--|
| `origin/master` — next 16.2.12, dev | ✅ passes |
| this branch — next 16.3.4, dev | ❌ `this.value.push is not a function` |
| this branch — next 16.3.4, `next build` + `next start` | ✅ renders, testid present |

**It's dev-only.** A production build serves the same page fine. The failure is in the Turbopack dev server's RSC payload for a *client-side* navigation — a hard `page.goto()` to the same URL server-renders correctly (200, full markup); only the Flight stream for the soft navigation fails to parse.

**Ruled out along the way:**
- `experimental.optimizePackageImports` on the icon barrels — disabling it appears to fix it only because `.next` was stale; with a genuinely clean cache it still fails.
- `withSentryConfig` / OTel server-component instrumentation — bypassing the wrapper entirely does not help, despite OTel frames showing in the 16.3.0 variant of the error.
- `@base-ui/react`, `motion`, and the rest of the sweep — this branch has none of them.
- `<GamificationHeaderCard />` in the dashboard layout — removing it does not help.

On 16.3.0 the same crash surfaces with a different message that names a chunk export (`Element type is invalid. Received a promise that resolves to: getUiState, …`), which is the same Flight-decode failure reported one layer higher.

### Suggested next step

Hold at 16.2.12. Worth a minimal reproduction against 16.3.x for an upstream issue — the interesting shape is "soft navigation to a large authenticated RSC page, Turbopack dev only".

#708 has been rebased off this branch so the patch + minor sweep can land independently.

guillermoscript:
Closing: Next.js 16.3.4 is a **NO-GO** and this branch has gone conflicting.

The blocker is a dev-only React Flight decode crash on soft-navigation to the student dashboard — reproduced on this branch, absent on 16.2.12. It is not something to work around in app code, so the branch has no path forward as-is.

Tracked in the dependency sweep notes; we retry on 16.4. The other two sweeps from the same batch shipped: #708 (34 patch/minor + both Sentry SDK warnings) and #711 (vitest 5, motion 13, mailgun.js 14).

PR #709, opened 2026-09-08
chore(deps): vitest 5, motion 13, mailgun.js 14 — the three unblocked majors

guillermoscript:
> Stacked on #708. Merge that first; this retargets to `master` automatically.

## What

`ncu` offers eight majors. These three are the only ones whose blockers are actually gone:

| package | from → to | note |
|--|--|--|
| `vitest` | 4.1 → 5.0 | **also adds `vite ^8` as a direct devDependency** — see below |
| `motion` | 12.43 → 13.2 | react 18/19 peers unchanged; all 10 call sites import from `motion/react` or `motion/react-client` |
| `mailgun.js` | 13.3 → 14.0 | reached only through `lib/email/client.ts`; engines `>=18` |

### vitest 5 unbundles vite

Vitest 5 no longer ships `vite`; it is a hard peer. `vite` was previously present only transitively, so installing vitest 5 removes it and `vitest run` dies at startup:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'vite' imported from node_modules/vitest/dist/chunks/…
```

Adding `vite@^8` to devDependencies fixes it. Node 26.7 satisfies vitest 5's `^22.12 || ^24 || >=26` engine range.

## The five majors still blocked, and by what

Checked against the registry rather than carried over from the last sweep's notes:

| major | blocker | still blocked? |
|--|--|--|
| `typescript` 5 → 7 | Next now ships `experimental.useTypeScriptCli`, so the `next build` half is solved — but `typescript-eslint` still needs the JS compiler API that TS 7 drops | yes, but it is now a spike rather than a wall |
| `eslint` 9 → 10 | `eslint-plugin-react@7.37.5` is the latest and still peers `eslint ^9.7`, and still calls the removed `context.getFilename()` | yes |
| `shiki` 3 → 4 | `@streamdown/code@1.1.1` is the latest and depends on `shiki ^3.19.0` | yes |
| `@solana/kit` 6 → 8 | `@solana/pay@1.0.26` peer-pins `@solana/kit ^6.9.0` | yes |
| `@solana/pay` 0.2 → 1.0 | peers satisfied by our kit 6.10, but it also requires `@solana-program/memo`, `@solana-program/token-2022` and `@solana/kit-plugin-instruction-plan`, none of which are installed. npm skips missing peers silently, so this fails at build time, not install time | needs its own PR |

## QA

- `npm run typecheck` — clean
- `npm run build` — compiled successfully
- `npm run test:unit` — 85 files / 1077 tests passing on vitest 5
- `npx playwright test --workers=1 gamification auth-security admin-pages email-mailer-state` — **40/40 passing** (gamification and admin dashboards are the motion-heavy surfaces; `email-mailer-state` covers the mailgun wrapper)

https://claude.ai/code/session_01A6ghiHrLpvdQpcFzaTNi4V

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -71,9 +71,9 @@
         "litesvm": "^1.4.1",
         "lru-cache": "^11.5.2",
         "lucide-react": "^1.43.0",
-        "mailgun.js": "^13.3.0",
+        "mailgun.js": "^14.0.1",
         "media-chrome": "^4.19.2",
-        "motion": "^12.43.0",
+        "motion": "^13.2.0",
         "nanoid": "^6.0.1",
         "next": "16.2.12",
         "next-intl": "^4.14.2",
@@ -118,7 +118,8 @@
         "tailwindcss": "^4",
         "tsx": "^4.23.13",
         "typescript": "^5",
-        "vitest": "^4.1.11"
+        "vite": "^8.2.2",
+        "vitest": "^5.0.0"
       }
     },
     "node_modules/@ai-sdk/gateway": {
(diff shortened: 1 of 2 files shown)


guillermoscript:
Superseded by #711 — this PR was auto-closed when its base branch was deleted by the #708 merge. Same commit, cherry-picked onto `master`.

PR #713, opened 2026-09-11
fix(#679): version-skew protection and narrow Sentry noise filters for the three open production issues

### LINKED ISSUE #679: fix: resolve the three open production Sentry issues (9F hooks order, 9M removeChild, 82 stale server action)
## Issues (environment:production, last 30 days)
- **LMS-FRONT-9F** `Rendered more hooks than during the previous render` at `/:locale/dashboard/student/profile` — a conditional hook in the profile page or a child.
- **LMS-FRONT-9M** `NotFoundError: Failed to execute 'removeChild' on 'Node'` at `/:locale/dashboard/student/courses/:courseId` — usually a browser extension or a DOM mutation outside React (video player / MDX embed). Confirm before chasing.
- **LMS-FRONT-82** `Failed to find Server Action` on `POST /[locale]/auth/login/page` — stale client after a deploy. Mitigation: reload-on-version-mismatch or a friendly retry on the login form.

## Acceptance
- [ ] Each issue either fixed with a regression test or resolved in Sentry with a one-line reason.

Part of #682.

guillermoscript:
## What

Wires Next.js version-skew protection (`deploymentId`) and adds two deliberately narrow Sentry drop rules for the production issues in #679. Two of the three are resolved with the reasoning recorded in code; the third (LMS-FRONT-9F) is **not fixed** — it gets a regression net instead, and why is spelled out below.

Closes #679

### LMS-FRONT-82 — `Failed to find Server Action` on `POST /[locale]/auth/login/page`

Reading Next 16.2's `server/app-render/action-handler.js` changes the diagnosis. There is exactly **one** path that throws this; every other one only warns:

| Request shape | Handling |
|---|---|
| fetch action (hydrated client, `next-action` header) with an unknown id | `handleUnrecognizedFetchAction()` → `console.warn` + `404` + `x-nextjs-action-not-found`. **Never throws.** |
| `application/x-www-form-urlencoded` POST, no action header | returns `null`, falls through to a normal page render. |
| **`multipart/form-data` POST, no action header, no valid `$ACTION_ID_`** | **throws** → 500 → Sentry. |

`getServerActionRequestMetadata()` marks *any* POST carrying a form content type as `isPossibleServerAction`, so an arbitrary multipart POST at a page URL enters the Server Action handler and dies there. `/auth/login` renders no Server Action at all (`components/login-form.tsx` calls `supabase-js` from an `onSubmit` handler; the layout adds none), so a genuine action POST cannot originate on that route — which leaves deploy skew on the routes that *do* have progressive-enhancement forms, and third parties POSTing form bodies at page URLs. Either way, the app had **no version-skew protection configured at all**.

### LMS-FRONT-9M — `NotFoundError: Failed to execute 'removeChild' on 'Node'`

Confirmed before chasing, as the issue asked. There is no in-app DOM mutation on `/dashboard/student/courses/[courseId]`: no `removeChild`/`insertBefore`/`appendChild`/`dangerouslySetInnerHTML` anywhere in that route's tree (the only direct DOM write in all of `components/` is `onboarding-checklist.tsx`'s clipboard textarea, which creates and removes its own detached node, and is not on this route). This is the React-vs-page-translator class: the translator rewrites React-owned text nodes into `<font>` wrappers, and React's next `removeChild` names a node that is no longer a child.

### LMS-FRONT-9F — `Rendered more hooks than during the previous render`

**Not fixed.** No conditional-hook pattern exists on the route, verified three ways:

1. A line-by-line read of every client component the page renders (`profile-form`, the five `gamification/*` components, `tours-toggle`, `student-certificate-card`, `social-share-modal`) and their custom hooks (`useGamificationSummary`, `useAchievements`, `useLeagueOptOut`, `useLeague`), plus the shared dashboard chrome — which matters, because the sibling route `/dashboard/student` reports the same error as LMS-FRONT-8F, so the common denominator is the shell.
2. A repo-wide `react-hooks/rules-of-hooks` run. All 12 findings are `render()` methods in `lib/puck/components/**` — Puck's lowercase-named component factories, false positives, and not on this route.
3. Base UI's `useRender`, used by the sidebar primitives on both routes, was checked directly: `useRenderElementProps` switches between `useMergedRefs` and `useMergedRefsN` and calls `void useMergedRefs(null, null)` in the disabled branch specifically to keep the hook count constant. Both helpers call exactly one hook. Ruled out.

This matches the repo's own prior triage — 9561748d (#632) deferred 9F and 8F as un-diagnosable without source maps. That upload wiring landed in the same commit, and nothing in this tree has changed since, so the next production occurrence should finally be symbolicated. Local reproduction was not possible in this session (Docker daemon down, so no local Supabase). Rather than guess at a fix, this PR adds a test that fails on the symptom.

## How to QA

On a fresh `npm run db:reset`, with the dev server on `lvh.me` (never `localhost`):

1. `npm run test:unit` — 87 files / 1097 tests pass, including the 12 new ones in `tests/unit/sentry-noise.test.ts`. Read that file first: it is the specification for both drop rules, including the cases that must **not** be dropped.
2. `SENTRY_RELEASE=deadbeefcafe npm run build`, then confirm the deployment id is baked in: `grep -o 'data-dpl-id="[^"]*"' .next/server/app/_not-found.html` prints `data-dpl-id="deadbeefcafe"`, and `grep -rl 'dpl=deadbeefcafe' .next/server/app` is non-empty (assets now carry `?dpl=`).
3. `npm run build` with `SENTRY_RELEASE` unset — `deploymentId` resolves to `undefined` and nothing changes, so local builds and local dev are unaffected.
4. `npx playwright test dashboard-client-errors --workers=1` — signs in as `student@e2etest.com` / `password123` on `http://lvh.me:3005`, loads `/en/dashboard/student` and `/en/dashboard/student/profile`, and fails if either emits a React hook-order crash. Expected to pass; it is a net for a bug that is not currently reproducible on demand.
5. Sanity-check the 9M keep path by hand: in the browser console on any dashboard page, run `Sentry.captureException(Object.assign(new Error("Failed to execute 'removeChild' on 'Node'"), { name: 'NotFoundError' }))`. With no translator active the event should still arrive in Sentry, tagged `dom.third_party_mutation: none-detected`. Right-click → Translate to Spanish first and the same call should produce no event at all.

## Screenshots / GIF

Not applicable — no user-visible surface changes. The diff is build config, two Sentry `beforeSend` predicates, and tests.

## Checklist

- [x] `npm run build` passes
- [x] Every new tenant-scoped query filters by `tenant_id` (or the table genuinely has no such column) — no queries added
- [ ] Tested with every relevant role (student / teacher / admin) — no role-dependent behaviour; the new Playwright spec covers the student routes only
- [x] Loading and error states handled — unchanged
- [x] Migration (if any) applies cleanly on `npm run db:reset`, has RLS policies, and `lib/database.types.ts` was regenerated — no migration

## Reviewer notes

- **The risk in this PR is the two `beforeSend` drops**, since a rule that is too broad hides a real crash silently. Both are pure predicates in `lib/sentry/noise.ts` rather than `ignoreErrors` substrings, both are pinned by tests that assert what is *kept* as well as what is dropped, and the 9M rule drops only with positive evidence of a third-party translator — otherwise it keeps the event and tags it. Both run **before** the OpenPanel pointer, so a dropped event never leaves an `error_captured` row pointing at an event Sentry never stored.
- `deploymentId` changes every asset URL on each deploy (`?dpl=<sha>`), which busts CDN and browser caches per release. That is the intended trade for skew protection, but it is a real behaviour change worth agreeing on.
- It resolves from `SENTRY_RELEASE` because that is where the commit SHA already is: `.dockerignore` excludes `.git`, so the SHA cannot be inferred inside the image, and `deploy.yml` already passes `github.sha` as a build arg the Dockerfile exports before `npm run build`. `NEXT_DEPLOYMENT_ID` overrides it and is documented in `.env.example`. **No Dokploy change is needed** — this is build-time only, and per `CLAUDE.md` the Dokploy service environment would be a silent no-op anyway.
- **LMS-FRONT-9F is being resolved in Sentry** with the one-line reason `#679`'s acceptance allows: *no conditional hook exists on this route (three independent audits); PR #713 lands a CI guard that fails on the symptom, so any recurrence re-opens this issue with a source-mapped stack.* Resolving rather than ignoring is the point — a resolved Sentry issue auto-reopens as a regression on the next event, and that event will finally carry a symbolicated stack, so nothing is lost. The `dashboard-client-errors` spec is the second net, in CI, before it can reach production again.

https://claude.ai/code/session_01U2i9mdjTkFajBYsAQa2Jaf

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -142,6 +142,15 @@ SENTRY_AUTH_TOKEN=                                 # Optional — without it `wi
                                                    # offsets), which is why browser issues cannot be traced back to a
                                                    # component. Generate at Sentry → Settings → Auth Tokens.
 
+# ─── VERSION SKEW ─────────────────────────────────────────────────────────────
+# Next.js `deploymentId`. Set it and Next stamps `?dpl=<id>` on every asset and
+# `x-deployment-id` on client navigations, so a browser still holding the last
+# deploy's bundle hard-navigates instead of firing requests the new server
+# cannot serve (LMS-FRONT-82, "Failed to find Server Action"). Build-time only.
+# It defaults to SENTRY_RELEASE — the commit SHA deploy.yml already passes in —
+# so there is normally nothing to set here; this exists to override that.
+NEXT_DEPLOYMENT_ID=                                # Optional — unique per deploy; unset falls back to SENTRY_RELEASE
+
 # ─── TESTING ──────────────────────────────────────────────────────────────────
 # Use lvh.me, not localhost: it resolves to 127.0.0.1 with wildcard subdomains, which
 # tenant routing needs. On localhost no subdomain resolves and every authenticated
--- a/instrumentation-client.ts
+++ b/instrumentation-client.ts
@@ -1,6 +1,7 @@
 import * as Sentry from "@sentry/nextjs";
 import { ANALYTICS_EVENTS } from "@/lib/analytics/events";
 import { shouldDropEvent } from "@/lib/analytics/exclusions";
+import { classifyDomMutationError } from "@/lib/sentry/noise";
 
 // The OpenPanel tracker installs a queueing `window.op` stub before the script
 // loads, so calling it here is safe at any point in the page lifecycle — but it
@@ -73,6 +74,21 @@ Sentry.init({
   // and joined back to the full Sentry event by id. Nothing here may ever block
   // or drop the Sentry event itself — the pointer is strictly best-effort.
   beforeSend(event) {
+    // LMS-FRONT-9M. A `removeChild` NotFoundError raised while a page
+    // translator is rewriting our text nodes is React losing a race with code
+    // we do not control; the same error with no translator present may be a
+    // real reconciliation bug, so that one is kept and tagged rather than
+    // dropped. Runs before the OpenPanel pointer so a dropped event never
+    // leaves an `error_captured` row pointing at nothing.
+    const domVerdict = classifyDomMutationError(
+      event,
+      typeof document === "undefined" ? null : document
+    );
+    if (domVerdict.kind === "drop") return null;
+    if (domVerdict.kind === "keep") {
+      event.tags = { ...event.tags, "dom.third_party_mutation": "none-detected" };
+    }
+
     try {
       const op = (window as { op?: OpenPanelGlobal }).op;
       if (
--- a/lib/build/deployment-id.ts
+++ b/lib/build/deployment-id.ts
@@ -1,0 +1,29 @@
+/**
+ * The build's deployment identifier, read at build time by `next.config.ts`.
+ *
+ * Next.js uses it for version-skew protection: it stamps `?dpl=<id>` on every
+ * static asset, sends `x-deployment-id` with client navigations and Server
+ * Action requests, and returns `x-nextjs-deployment-id` on navigation
+ * responses. A client still holding the previous deployment's bundle sees the
+ * mismatch and performs a hard navigation instead of firing requests the new
+ * deployment cannot serve — which is the condition behind LMS-FRONT-82
+ * (`Failed to find Server Action`, issue #679).
+ *
+ * Nothing about this is Sentry-specific; `SENTRY_RELEASE` is simply where the
+ * commit SHA already lives. `.github/workflows/deploy.yml` passes `github.sha`
+ * as a Docker build arg and the Dockerfile exports it before `npm run build`,
+ * so it is the one value that is (a) unique per deploy and (b) already present
+ * at build time. `.dockerignore` excludes `.git`, so the SHA cannot be inferred
+ * from the repo inside the image. `NEXT_DEPLOYMENT_ID` — Next's own convention
+ * — wins when set, so an operator can override without touching Sentry.
+ *
+ * Unset (local dev, `npm run build` on a laptop) resolves to `undefined`, which
+ * is exactly "no deployment id" and leaves Next's behaviour unchanged.
+ */
+export function resolveDeploymentId(
+  env: Record<string, string | undefined> = process.env
+): string | undefined {
+  const candidate = env.NEXT_DEPLOYMENT_ID ?? env.SENTRY_RELEASE
+  const trimmed = candidate?.trim()
+  return trimmed ? trimmed : undefined
+}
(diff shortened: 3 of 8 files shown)


guillermoscript:
Merged as `fb847d3b` with no review changes — what shipped is exactly what the description above covers. All seven checks were green at merge, including all four E2E shards.

Final state of the three production issues from #679:

- **LMS-FRONT-82** — resolved two ways. `deploymentId` now resolves from the build SHA (`lib/build/deployment-id.ts`, read by `next.config.ts`), which is the first version-skew protection this app has had; and the error itself is dropped server-side via `isServerActionNotFoundError`. The diagnosis is the part worth remembering: Next 16.2 throws this from exactly one path — a `multipart/form-data` POST with no `next-action` header and no resolvable `$ACTION_ID_` — while the stale-fetch-action case only warns and 404s. `/auth/login` renders no Server Action at all, so what was hitting it was form-shaped POSTs we do not originate.
- **LMS-FRONT-9M** — dropped only when a page-translator marker is actually present in the document; with none, the event is kept and tagged `dom.third_party_mutation: none-detected`, so a genuine React reconciliation bug still reaches us. Confirmed first, as the issue asked, that no in-app DOM mutation exists anywhere on the course route.
- **LMS-FRONT-9F** — **not fixed in code, and resolved in Sentry instead** (done by @guillermoscript after this merge), which is the second branch of #679's acceptance criteria. Three independent audits found no conditional hook on the route: every client component and custom hook it renders plus the shared dashboard chrome, a repo-wide `react-hooks/rules-of-hooks` run whose only hits are Puck's lowercase `render()` factories, and Base UI's `useRender`, which holds its hook count constant by construction. A resolved Sentry issue auto-reopens as a regression on the next event, and that event will carry a source-mapped stack — which is what this bug has been waiting on since #632 deferred it. `tests/playwright/dashboard-client-errors.spec.ts` is the CI net in the meantime; it ran and passed on shard 2 of this PR's E2E job.

Verification: `npm run typecheck`, `npm run test:unit` (87 files / 1097 tests, 12 new), `npm run build` with `SENTRY_RELEASE` set — confirming `?dpl=` on assets and `data-dpl-id` on `<html>` — and `Production image builds` green in CI, which exercises the same config change through the real Docker pipeline.

Nothing was deferred to a follow-up issue.

Issue #676, opened 2026-09-04
fix(email): stop pretending emails were sent — surface Mailgun state in the UI and configure it in prod


guillermoscript:
## Why
All app emails go through `lib/email/send.ts`, which does `console.warn(...)` and `return false` when `MAILGUN_API_KEY` / `MAILGUN_DOMAIN` are unset, and also swallows Mailgun API errors. Only `access-cutoff.ts` checks the boolean. Mailgun is **not configured in prod**, so invitations, certificate emails, welcome emails and plan reminders send nothing while the UI shows success.

## Scope
- App: every call site that reports to a user (invitations, certificate issue, join-school welcome) must read the boolean and say "Email not sent — share this link instead" when false.
- Admin: a one-line status in Settings → Email showing whether the platform mailer is configured (read-only, from env presence).
- Ops (user-only): set `MAILGUN_API_KEY`, `MAILGUN_DOMAIN`, `EMAIL_FROM` in Dokploy runtime env, verify one invitation arrives.
- `app/actions/teacher/courses.ts:420` course-deletion email: move to a template, i18n, batch the `getUserById` N+1.

## Acceptance
- [ ] With Mailgun unset locally, inviting a user shows the "not sent" state and the copyable link.
- [ ] With Mailgun set in prod, one real invitation email received.

Part of #682.

guillermoscript:
> An AI agent (Claude Code, driven by @guillermoscript) is picking this issue up. Plan below; the PR will follow the same outline.

## Root cause / current behavior

- `lib/email/send.ts` is the only mailer. When `MAILGUN_API_KEY` or `MAILGUN_DOMAIN` is unset it logs a `console.warn` and returns `false`; a Mailgun API error also returns `false`. Nothing distinguishes "not configured" from "failed".
- Ten call sites use it. Only three of them report to a person in the UI, and only one reads the boolean:
  - **Invitations** (`app/actions/admin/invitations.ts`): fixed in #695 — the dialog already shows "Invitation created — email not sent" plus the join link.
  - **Certificate issue** (`app/api/certificates/issue/route.ts`, both the signed and the simplified path): the boolean is discarded, the route returns `{ success, certificateId }`, and `IssueCertificateButton` toasts "Certificate issued to {name}" — the teacher assumes the student got the email.
[...]

PR #715, opened 2026-09-11
fix(certificates): lessons-only courses can earn a certificate again (#696)

### LINKED ISSUE #696: fix(certificates): lessons-only courses can never earn a certificate — calculate_course_completion returns eligible: null when a course has no exams
## Why
`calculate_course_completion(p_user_id, p_course_id)` decides exam eligibility with

```sql
(v_template.requires_all_exams = true AND v_submitted_exams = v_total_exams AND v_all_exams_passed)
```

where `v_all_exams_passed` is `BOOL_AND(esc.score >= min_exam_pass_score)` over the student's scored submissions. For a course with **zero published exams** there are no rows, `BOOL_AND` returns `NULL`, and the whole predicate collapses to `NULL` → `eligible: null`, which every caller treats as "not eligible". The `requires_all_exams = false` branch fails too, because `AVG()` over no rows is `NULL`.

Net effect: a course with lessons but no exams can never issue a certificate, even at 100% lesson completion with an active template. Observed locally while recording QA for #676 (course 1002 "Web Development Basics", 1/1 lessons, active template):

```json
{"eligible": null, "totalExams": 0, "totalLessons": 1, "completedLessons": 1, "completionPercentage": 100.00, "allExamsPassed": false, "submittedExams": 0}
```

## Fix
Treat "no exams" as passed: `COALESCE(v_all_exams_passed, true)` when `v_total_exams = 0` (and skip the average-score test in the `requires_all_exams = false` branch when there are no exams). Add a regression test for a lessons-only course.

Function last touched in `supabase/migrations/20260216000100_create_certificates_system.sql`; `20260906120000_certificate_issue_on_exam_scores.sql` only adds the trigger that calls it.

guillermoscript:
## What

`calculate_course_completion()` returned `"eligible": null` for any course with no published exams, so a course made of lessons alone could never issue a certificate. A new migration decides the exam requirement in an explicit branch and treats "no published exams" as vacuously satisfied, and a rolled-back SQL regression script pins the behaviour. One follow-on: the Open Badges evidence string now omits its exam clause for a course with no exams, instead of baking "0% average on 0 exams" into a permanent credential.

Closes #696

## Why

The function decided eligibility inside one boolean expression:

```sql
v_eligible := (
  v_completion_pct >= v_template.min_lesson_completion_pct
  AND (
    (v_template.requires_all_exams = true AND v_submitted_exams = v_total_exams AND v_all_exams_passed)
    OR (v_template.requires_all_exams = false AND v_avg_score >= v_template.min_exam_pass_score)
  )
);
```

`v_all_exams_passed` is `BOOL_AND(...)` and `v_avg_score` is `AVG(...)` over the student's scored submissions. With zero published exams both aggregates run over an empty set and return `NULL`, and in three-valued logic `true AND NULL` is `NULL` and `NULL OR NULL` is `NULL` — so the verdict was neither true nor false. `jsonb_build_object` emitted it faithfully as `"eligible": null`, and every consumer reads that as "not eligible": `check_and_issue_certificate` casts it with `(v_completion->>'eligible')::BOOLEAN`, and the TypeScript helpers in `lib/certificates/issue-certificate.ts` use `data.eligible || false`.

Because both auto-issuance triggers (`lesson_completions` and `exam_scores`) route through this function, the entire certificate automation was dead for lessons-only courses — at any completion percentage, with any active template. Exactly the payload reported in the issue, reproduced on `master`:

```
ERROR:  §3 eligible is null — this is the #696 regression (payload: {"eligible": null,
  "totalExams": 0, "totalLessons": 2, "completedLessons": 2, "completionPercentage": 100.00,
  "allExamsPassed": false, "submittedExams": 0, ...})
```

### What the migration changes

- **"No published exams" satisfies the exam requirement.** Decided in an `IF`/`ELSIF` ladder over a dedicated `v_exams_ok` variable *before* the `requires_all_exams` branch, so it covers both template modes at once rather than patching each separately.
- **The nullable template thresholds are read through `COALESCE`** to their table defaults (100 / 70 / true). `min_lesson_completion_pct`, `min_exam_pass_score` and `requires_all_exams` all have defaults but are nullable, so an explicit `NULL` in any of them produced the identical `eligible: null` symptom by a different route.
- **The verdict itself is `COALESCE`d to `false`**, so the function can never again return a null eligibility regardless of what a future edit does upstream of it.
- **A course with no published lessons *and* no published exams is explicitly not eligible.** Without this guard, "no exams means passed" plus a template whose `min_lesson_completion_pct` is `0` would hand out certificates for a completely empty course. That case returns `null` today (read as not eligible), so the guard preserves the current outcome rather than changing it.

The result JSON keeps its shape and keys. The only differences are that `eligible` is now always a real boolean, `allExamsPassed` reports `true` rather than `false` when there are no exams to pass, and `reason` accompanies the explicit not-eligible early returns.

The replacement keeps the original's `STABLE` / `SECURITY INVOKER` attributes (verified against `pg_proc`, not just the migration text). Nothing in the application duplicates the verdict — `lib/certificates/issue-certificate.ts` and the MCP `certificates` tool both delegate entirely to this RPC.

### The one application-code consequence

Making lessons-only courses certifiable makes `generateAchievementCredential()` in `lib/certificates/open-badges.ts` reachable for a course with no exams, where its evidence string read:

> Completed 1 of 1 lessons (100%) **and achieved 0% average on 0 exams**

That text is baked into a permanent Open Badges credential and served from the public `/verify/<code>` page, so the exam clause is now omitted entirely when `totalExams` is 0. `tests/unit/certificate-credential-evidence.test.ts` covers both shapes.

### Deliberately out of scope

- **No retroactive backfill.** Students who already finished a lessons-only course get nothing until a trigger fires for them again. Issuing certificates retroactively is a data migration with its own blast radius and belongs in its own change.
- `required_lesson_ids` / `required_exam_ids` on the template are still unread by this function, as they always have been.
- Heads-up on a doc drift found while surveying, not fixed here: `docs/DATABASE_SCHEMA.md` (lines 671 and 891) describes `calculate_course_completion()` and `check_and_issue_certificate()` as `SECURITY DEFINER`. Neither ever was — `pg_proc.prosecdef` is `false` for both. Only `issue_certificate_if_eligible()` is a definer function.

## How to QA

The regression lives at the database level, so the script is the QA — it builds its own fixtures, asserts, and rolls everything back, leaving the seeded database untouched.

1. `npm run db:reset`
2. Run the regression script:
   ```bash
   docker exec -i supabase_db_lms-front psql -U postgres -d postgres -P pager=off \
     < tests/sql/issue-696-lessons-only-certificate.sql
   ```
   Expected: `NOTICE: PASS: #696 — lessons-only courses are certificate-eligible, exam courses and empty courses unchanged`, followed by `ROLLBACK`.
3. To see the bug it fixes, check out `master` (or just run the script *before* applying `supabase/migrations/20260911120000_lessons_only_certificate_eligibility.sql`) and run the same script. It fails at `§3` with the `eligible: null` payload quoted above.

The eight assertions, in order:

| § | Case | Expected |
|---|---|---|
| 1 | Lessons-only course, nothing completed | `eligible` is a **boolean**, and `false` |
| 2 | Lessons-only course, 1 of 2 lessons done | `false`, `completionPercentage` 50 |
| 3 | **The regression** — lessons-only, 2 of 2 done | `true` (was `null`) |
| 4 | The `lesson_completions` trigger chain | exactly one certificate row auto-issued |
| 5 | Template with NULL thresholds | boolean verdict, falls back to the defaults |
| 6 | Course *with* an exam: unsubmitted, then failed | `false` both times — unchanged behaviour |
| 7 | Course with an exam: passed, lessons done | `true` — unchanged behaviour |
| 8 | Course with nothing published, 0% threshold | `false` — the new guard |

§4 is worth a look: it asserts the real trigger chain (`lesson_completions` → `on_lesson_completed_trigger` → `issue_certificate_if_eligible` → `check_and_issue_certificate` → this function) actually writes a `certificates` row, which is the user-visible symptom from the #676 QA recording.

For the credential evidence string, `npm run test:unit` is the check — `tests/unit/certificate-credential-evidence.test.ts` asserts both shapes (no exams: no exam clause; exams: clause intact).

There is no seeded course affected by this locally — the local seed has no `certificate_templates` rows at all, which is why the bug survived this long.

## Screenshots / GIF

None — this is a database function plus one string inside the machine-readable Open Badges credential. Neither has a rendered surface: no component or page reads the credential's `evidence` block (the `/verify/<code>` page renders `completion_data`, not the credential JSON). The verification evidence is the before/after script run quoted above.

## Checklist

- [x] `npm run typecheck` and `npm run test:unit` pass — typecheck clean, 1099 tests across 88 files green
- [x] `npm run build` passes — clean, exit 0
- [x] Every new tenant-scoped query filters by `tenant_id` (or the table genuinely has no such column) — no new queries; the function's existing reads are scoped by `course_id`/`user_id` as before
- [x] Tested with every relevant role (student / teacher / admin) — the function is role-agnostic; the script exercises it for a student, which is the only role that earns certificates
- [x] Loading and error states handled — n/a, no UI
- [x] New UI strings added to both `messages/en.json` and `messages/es.json` — n/a, no new UI strings (the credential evidence text is machine-readable Open Badges payload, untranslated today as before)
- [x] Migration applies cleanly, has RLS policies, and `lib/database.types.ts` was regenerated — applied cleanly to the fully-migrated local database; it is a standalone `CREATE OR REPLACE FUNCTION` with no ordering dependency, so no RLS policy is involved and the RPC signature is unchanged, leaving `lib/database.types.ts` correct as-is (and `npm run db:types` reads `--linked`, i.e. cloud, so regenerating it here would pull unrelated drift)

> **After merge:** this migration still needs `supabase db push` to reach the cloud database — it is not applied there yet.

https://claude.ai/code/session_01Qd5dVvZZAWKjJfqQwFajUU

### DIFF
--- a/lib/certificates/open-badges.ts
+++ b/lib/certificates/open-badges.ts
@@ -163,7 +163,14 @@ export function generateAchievementCredential(
             id: `${appUrl}/verify/${verificationCode}`,
             type: ['Evidence'],
             name: 'Course Completion Evidence',
-            description: `Completed ${completionData.completedLessons} of ${completionData.totalLessons} lessons (${completionData.completionPercentage}%) and achieved ${completionData.averageExamScore}% average on ${completionData.submittedExams} exams`,
+            // Courses with no exams are certifiable (#696), so the exam clause is
+            // omitted rather than baking "0% average on 0 exams" into a permanent
+            // credential.
+            description:
+                `Completed ${completionData.completedLessons} of ${completionData.totalLessons} lessons (${completionData.completionPercentage}%)` +
+                (completionData.totalExams > 0
+                    ? ` and achieved ${completionData.averageExamScore}% average on ${completionData.submittedExams} exams`
+                    : ''),
             genre: 'Certificate',
         },
     ];
--- a/supabase/migrations/20260911120000_lessons_only_certificate_eligibility.sql
+++ b/supabase/migrations/20260911120000_lessons_only_certificate_eligibility.sql
@@ -1,0 +1,180 @@
+-- A course with lessons but no exams could never earn a certificate (#696).
+--
+-- `calculate_course_completion()` decided eligibility with a single boolean
+-- expression:
+--
+--     v_completion_pct >= min_lesson_completion_pct
+--     AND (
+--       (requires_all_exams = true AND v_submitted_exams = v_total_exams AND v_all_exams_passed)
+--       OR (requires_all_exams = false AND v_avg_score >= min_exam_pass_score)
+--     )
+--
+-- `v_all_exams_passed` is `BOOL_AND(...)` and `v_avg_score` is `AVG(...)` over
+-- the student's scored submissions. For a course with zero published exams both
+-- aggregates run over an empty set and return NULL, so in three-valued logic the
+-- whole expression collapses to NULL — not false, but not true either. The
+-- function then emitted `"eligible": null`, and every caller
+-- (`check_and_issue_certificate` casting with `::BOOLEAN`, the TypeScript
+-- helpers using `data.eligible || false`) reads that as "not eligible". A
+-- lessons-only course was therefore un-certifiable at 100% lesson completion
+-- with an active template, and both auto-issuance triggers were dead for it.
+--
+-- Fix: decide the exam requirement in an explicit branch instead of inside the
+-- boolean expression, and treat "no published exams" as vacuously satisfied.
+-- The nullable template thresholds are read through COALESCE to their documented
+-- defaults for the same reason — an explicit NULL in `min_lesson_completion_pct`,
+-- `min_exam_pass_score` or `requires_all_exams` produced the identical
+-- `eligible: null` symptom by a different route — and the verdict itself is
+-- COALESCEd so the function can never again return a null eligibility.
+--
+-- A course with no published lessons AND no published exams is now explicitly
+-- not eligible. Without that guard, "no exams means passed" plus a template
+-- whose `min_lesson_completion_pct` is 0 would issue certificates for a
+-- completely empty course; today that case returns null (read as not eligible),
+-- so the guard preserves the current outcome rather than changing it.
+--
+-- The result JSON keeps its shape: same keys, `eligible` now always a real
+-- boolean, and `reason` present on the explicit not-eligible early returns.
+
+CREATE OR REPLACE FUNCTION public.calculate_course_completion(
+  p_user_id UUID,
+  p_course_id INTEGER
+)
+RETURNS JSONB
+LANGUAGE plpgsql
+STABLE
+AS $$
+DECLARE
+  v_total_lessons INTEGER;
+  v_completed_lessons INTEGER;
+  v_completion_pct NUMERIC;
+  v_total_exams INTEGER;
+  v_submitted_exams INTEGER;
+  v_avg_score NUMERIC;
+  v_all_exams_passed BOOLEAN;
+  v_exams_ok BOOLEAN;
+  v_eligible BOOLEAN := false;
+  v_min_lesson_pct INTEGER;
+  v_min_exam_score INTEGER;
+  v_requires_all_exams BOOLEAN;
+  v_template RECORD;
+  v_result JSONB;
+BEGIN
+  -- Get certificate template for course
+  SELECT * INTO v_template
+  FROM public.certificate_templates
+  WHERE course_id = p_course_id AND is_active = true
+  LIMIT 1;
+
+  -- If no template, return not eligible
+  IF NOT FOUND THEN
+    RETURN jsonb_build_object(
+      'eligible', false,
+      'reason', 'No active certificate template for this course'
+    );
+  END IF;
+
+  -- The criteria columns are nullable; fall back to the table defaults so a NULL
+  -- can never make the eligibility verdict itself NULL.
+  v_min_lesson_pct := COALESCE(v_template.min_lesson_completion_pct, 100);
+  v_min_exam_score := COALESCE(v_template.min_exam_pass_score, 70);
+  v_requires_all_exams := COALESCE(v_template.requires_all_exams, true);
+
+  -- Count total published lessons
+  SELECT COUNT(*) INTO v_total_lessons
+  FROM public.lessons
+  WHERE course_id = p_course_id AND status = 'published';
+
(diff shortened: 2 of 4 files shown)


guillermoscript:
Merged as `729f1c20` (squash), closing #696. All seven checks were green at merge — `verify`, `Production image builds`, `MCP server unit tests`, and all four E2E shards — and the PR merged as reviewed, with no changes requested during review.

**What shipped**

- `supabase/migrations/20260911120000_lessons_only_certificate_eligibility.sql` replaces `calculate_course_completion()`. The exam requirement is now decided in an explicit `IF`/`ELSIF` ladder over a `v_exams_ok` variable instead of inside the eligibility expression, and zero published exams satisfies it vacuously. The three nullable template thresholds are read through `COALESCE` to their table defaults, and the verdict itself is `COALESCE`d, so the function can no longer return `"eligible": null` by any route. A course with no published lessons *and* no published exams is explicitly not eligible.
- `lib/certificates/open-badges.ts` omits the exam clause from the Open Badges evidence string when the course has no exams, instead of writing "0% average on 0 exams" into a permanent credential.
- `tests/sql/issue-696-lessons-only-certificate.sql` — eight assertions in a rolled-back transaction, including the trigger chain actually writing a `certificates` row. It fails at `§3` on the pre-merge function with the exact payload from the issue.
- `tests/unit/certificate-credential-evidence.test.ts` — the evidence string with and without exams.

**Still outstanding for whoever deploys this**

The migration is on `master` but **not yet applied to the cloud database** — it needs `supabase db push`. Until then, production still returns `eligible: null` for lessons-only courses.

**Noted, not fixed** (raised in the PR description, left for a separate change):

- `docs/DATABASE_SCHEMA.md` lines 671 and 891 describe `calculate_course_completion()` and `check_and_issue_certificate()` as `SECURITY DEFINER`. Neither ever was — `pg_proc.prosecdef` is `false` for both, and only `issue_certificate_if_eligible()` is a definer function. Worth correcting, because `CREATE OR REPLACE FUNCTION` silently resets that attribute when it isn't restated.
- No retroactive backfill: students who already completed a lessons-only course get nothing until a trigger fires for them again.
- `required_lesson_ids` / `required_exam_ids` on the template remain unread by this function, as they always have been.

Issue #684, opened 2026-09-06
ux: paid-course CTA loses checkout intent for a new visitor (join-school drops the URL)


guillermoscript:
## Problem

A first-time visitor who clicks **Enroll Now** on a paid course's public page (`/courses/<id>`) loses their purchase intent before they ever see the checkout.

The CTA links to `/auth/login?next=/checkout?courseId=<id>`. After the visitor signs up (or logs in) they are a member of no school yet, and:

1. `/checkout` is not a public route, so `proxy.ts` (`:432-444`) bounces the non-member to `/join-school` **without** preserving the original URL — no `next`, no `redirectTo`.
2. `joinCurrentSchool()` (`app/actions/join-school.ts`) takes no return target, and `join-school-form.tsx:41` always hard-navigates to `/dashboard/student` on success.

So the visitor ends up on an empty student dashboard and has to find the course again from scratch. The **free** path does not have this problem only because `enrollFree()` joins the school itself and `/courses` is public.

## Expected

After joining the school, the visitor lands on the checkout for the course they clicked (or at least back on the course page).

## Suggested fix

- Proxy: when redirecting a logged-in non-member to `/join-school`, append `?next=<original path + query>` (sanitised by `getSafeNextPath`).
- `join-school` page/form: read `next`, pass it through `joinCurrentSchool({ next })` or handle it client-side, and `window.location.assign(next || '/dashboard/student')` after a successful join.
- Optionally treat `/checkout` like `/courses` for members-to-be: let `enrollFree`'s "join on demand" pattern also apply to the checkout server actions.

## Context

Found while writing the Loop 2 E2E journey (#671). The Playwright spec there covers the free path end to end; the paid path is not covered until this is fixed.

Part of #682.

guillermoscript:
🤖 An AI agent (Claude Code, driven by @guillermoscript) is picking this issue up. Plan of attack below; the PR will follow the same outline.

## Root cause / current behavior

The paid CTA on `/courses/<id>` links to `/auth/login?next=/checkout?courseId=<id>`, and the auth pages do carry `next` through login, sign-up, `/auth/confirm` and `/api/auth/callback`. The intent is lost one hop later:

1. `/checkout` is not in `publicRoutes`, so for a logged-in user with no active `tenant_users` row `proxy.ts` (the non-member branch after the membership lookup) redirects to `/${locale}/join-school` with an empty query. Nothing records where the user was going.
[...]

Issue #671, opened 2026-09-04
test(e2e): Loop 2 — a student arrives by public link, joins, learns and downloads a verifiable certificate


guillermoscript:
## Why
`full-student-journey.spec.ts` covers lesson → exam → AI grading → certificate, but both ends are missing: entering by public link, sign-up + `/join-school` happy path, PDF download, and `/verify` with real data (3 of 5 verification tests self-skip). Student onboarding has no spec at all. Password reset has none either.

## Journey (one spec, `tests/playwright/loop-2-student-learns.spec.ts`)
1. Anonymous opens `https://code-academy.lvh.me/courses/<id>` → click enroll → redirected to sign-up with `next` preserved → sign up → `/join-school` → becomes member → lands on the course.
2. Free enrollment (or plan-covered) → lesson → complete → exercise submit (un-skip the two skipped tests) → exam → results.
3. Certificate appears on `/dashboard/student/certificates` → **Download PDF** produces a PDF response → **Verify** opens `/verify/<code>` anonymously and shows the student's name.
4. Log out → "Forgot password" → request → follow the local Inbucket link → `/auth/update-password` → log back in.

Mobile project (Pixel 5) must pass steps 1–3.

## Acceptance
- [ ] Spec passes in CI on desktop and mobile.
- [ ] `certificate-verification.spec.ts` no longer self-skips (seed a certificate in the spec).
- [ ] UI gaps filed as `ux` issues and linked.

Part of #682.

guillermoscript:
🤖 An AI agent (Claude Code, on behalf of @guillermoscript) is picking this up. Plan below; the PR will follow the same structure.

## Current behaviour

- `full-student-journey.spec.ts` is `test.fixme` — it depends on course 9999 / exam 9999 / lessons 10000–10001, which no seed file creates.
- `certificate-verification.spec.ts` only seeds a certificate when the table is empty, and the three "valid code" tests `test.skip()` when `existingVerificationCode` is unset — so on a fresh CI database they never run.
- `evaluations-security.spec.ts` keeps two `exercise_completions` tests skipped with a stale reason ("trigger references `l.lesson_id`" — fixed in `20260611120000_fix_xp_trigger_lesson_id.sql`). The real blocker is that the seeded `[E2E] Essay Exercise` has no `lesson_id`, so `handle_exercise_completion_xp` cannot resolve a tenant.
[...]

PR #718, opened 2026-09-12
fix(public): course CTA sends first-time visitors to sign-up, not login (#685)

### LINKED ISSUE #685: ux: public course CTA sends first-time visitors to login instead of sign-up
## Problem

The public course page (`/courses/<id>`) sends an anonymous visitor to the **login** page — `/auth/login?next=…` (`app/[locale]/(public)/courses/[id]/page.tsx:526`) — for both "Enroll for Free" and "Enroll Now".

The visitor most likely to click that button has never had an account (they arrived by a shared public link), so the very first thing they see is a form they cannot fill, and they have to spot the small "Don't have an account? Sign up" footer link to continue. The `next` parameter does survive that extra hop (the login page forwards it to sign-up), so nothing breaks — it is one avoidable step at the moment of highest intent.

## Expected

Either:
- send anonymous visitors to `/auth/sign-up?next=…` (with the login link as the secondary path on that page), or
- render two clear actions on the pricing card: **Create account** (primary) and **I already have an account** (secondary), both carrying `next`.

## Notes

- Keep `next` exactly as today (`/courses/<id>?enroll=1` for free, `/checkout?courseId=<id>` for paid) so `AutoFreeEnrollButton` keeps working.
- `sign-up-form.tsx` already reads `next` (`:50`) and pushes it after a live session.
- Found while writing the Loop 2 E2E journey (#671); the spec currently drives the login → sign-up hop explicitly.

Part of #682.

guillermoscript:
Closes #685.

## What was wrong

`/courses/<id>` pointed both "Enroll for Free" and "Enroll Now" at `/auth/login?next=…`. The visitor most likely to click that button arrived by a shared public link and has **never had an account**, so at the moment of highest intent they met a form they could not fill, and had to spot the small "Don't have an account? Sign up" footer link to continue. Nothing broke — `next` survived the hop — it was one avoidable step in the wrong place.

## What changed

- Primary button → `/auth/sign-up?next=…`.
- A secondary **"Already have an account? Log in"** sits under it, carrying the same `next`, so the returning student loses nothing.
- The two destinations are hoisted into one `anonymousNext` const (`/courses/<id>?enroll=1` when free so `AutoFreeEnrollButton` still finishes the job, `/checkout?courseId=<id>` when paid) — unchanged values, but the two links can no longer drift apart.
- New keys `coursePublicDetails.pricing.haveAccount` / `.logIn` in **en** and **es**.

`sign-up-form.tsx` already reads `next` (via `getSafeNextPath`) and forwards it through the confirm and OAuth callback URLs, so nothing else needed touching.

## Tests

Two specs drove the old login → sign-up hop explicitly and now expect the direct path — each also asserts the secondary login link carries the same intent, which preserves what the old hop covered:

- `loop-2-student-learns.spec.ts` (#671) — free course, `next=/courses/<id>?enroll=1`
- `paid-course-checkout-intent.spec.ts` (#684) — paid course, `next=/checkout?courseId=<id>`

```bash
npm run typecheck    # clean
npm run test:unit    # 1099 passed, incl. en/es catalogue parity
npx eslint …         # 0 errors (2 pre-existing <img> warnings on this file)
```

The local Supabase stack was down, so the two specs were not run on this machine — they run on every PR in CI.

## Noticed, deliberately not in this PR

Two other public CTAs have the same first-visitor shape and are **not** touched here, to keep this reviewable against #685's scope — filed separately:

- `app/[locale]/(public)/products/[productId]/page.tsx:98` — anonymous visitors get a **Login** button; the copy there is also hardcoded English ("Please login to request payment information").
- `components/public/free-subscribe-button.tsx:52` — the free-plan CTA on `/pricing` goes to login with `next=/pricing?subscribe=<planId>`.

The server-side `redirect()`s in `/checkout`, `/checkout/manual`, `/checkout/success` and `/join-school` are guards on already-intentful flows rather than entry CTAs, and are fine as login.

https://claude.ai/code/session_01GsCUfngsjLWq1gqVXVNaVj

### DIFF
--- a/app/[locale]/(public)/courses/[id]/page.tsx
+++ b/app/[locale]/(public)/courses/[id]/page.tsx
@@ -198,6 +198,14 @@ export default async function CourseDetailsPage(props: {
         ? t('pricing.free')
         : formatPrice(Number(courseProduct.price), courseProduct.currency);
 
+    // Where an anonymous visitor lands once they have an account. Unchanged
+    // from when the CTA pointed at login: the free path re-enters this page
+    // with `enroll=1` so `AutoFreeEnrollButton` finishes the job, the paid one
+    // goes straight to checkout.
+    const anonymousNext = isFree
+        ? `/courses/${params.id}?enroll=1`
+        : `/checkout?courseId=${course.course_id}`;
+
     const instructor = author ? {
         name: author.full_name || t('sections.instructor.defaultName'),
         avatar_url: author.avatar_url,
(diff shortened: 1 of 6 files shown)


guillermoscript:
CI shard 2 went red on the first push and the fix is in `ec46c9a9`.

`loop-1-creator-publishes.spec.ts` also pins the anonymous CTA's href — twice, once in the English pass and once in the Spanish one — and I had only updated `loop-2` and `paid-course-checkout-intent`. Both now expect `/auth/sign-up?next=` and additionally assert `course-enroll-login` carries its own `next`, so all three specs check the same contract.

One unrelated flake in the same shard, green on retry: `loop-3-student-pays.spec.ts:622` (a logged-in dashboard course heading, nothing to do with the public CTA).

PR #742, opened 2026-09-13
fix: pin the transactions trigger's search_path, and prove four money rails end to end

guillermoscript:
Closes #741

## What this fixes

**1 — `public.trigger_manage_transactions()` had no `SET search_path`**, so it ran under the *caller's*, and every reference in its body was unqualified.

Fine while every writer was a Node service-role client. Broken once the hardened SQL functions arrived, because those pin `SET search_path TO ''` and the trigger inherits it:

- `apply_webhook_refund` → **every refund** on `/api/payments/webhook/{lemonsqueezy,paypal,binance}` returned 500. Provider retries forever, `refunded_amount` never written, entitlement never revoked, `getPayoutsOwed` keeps counting a refunded sale.
- `settle_expired_checkout` → late settlement of an expired hosted checkout failed the same way.

The trigger's *first* statement runs for any row with a non-null `product_id` regardless of status, so this was never limited to the `successful` branch. Stripe and manual are unaffected — `/api/stripe/webhook` handles `charge.refunded` inline on a normal search_path.

Pinning it surfaced its last unpinned callee: `cancel_subscription(uuid, integer)` also had `proconfig = NULL` and a bare `UPDATE subscriptions`, so it would then have broken every plan-shaped transaction reaching `failed`. Both are now pinned and schema-qualified.

**2 — a student's course purchase could pay the school's platform bill.** `dispatchPlatformBillingEvent` had no list of the event types it models, so `STATUS_BY_TYPE[type] ?? 'active'` treated the student loop's `payment.succeeded` as an activation.

Reachable because Stripe is the only rail with a second signing secret (`STRIPE_PLATFORM_WEBHOOK_SECRET`): PayPal, Binance Pay and Lemon Squeezy sign **both loops with one secret, from one merchant account, under one webhook registration**. Register `/api/billing/webhook/<provider>` — which #610 tells operators to do for Binance, and which the PayPal endpoint invites by existing — and it also receives every student's course purchase, correctly signed and therefore verified. `resolveTenantId` reads `metadata.tenantId`, exactly the key the student checkout writes, so the event resolved the **buyer's school**; the identity guard below it is skipped whenever the stored row has no `provider_subscription_id`, which is the shape of every bank-transfer subscription.

[...]

PR #751, opened 2026-09-14
fix: settle manual sales on the request's school, label them manual, keep $0 items off the request form (#745 #746 #749)

### LINKED ISSUE #745: completeAndEnroll writes the acting admin's tenant, not the request's
From the PayPal + manual audit (PR #742).

## The inconsistency

`app/actions/payment-requests.ts` → `completeAndEnroll` (line 475) reads the request with a **super-admin bypass**:

```ts
if (fetchError || !request || (request.tenant_id !== tenantId && !superAdmin)) throw …
```

…and then writes with the **acting** tenant:

- the transaction insert uses `tenant_id: tenantId` (line 520) — the tenant of the subdomain the admin is on, not `request.tenant_id`;
- the request update filters `.eq('tenant_id', tenantId)` with no `.single()`, so when it matches nothing it **succeeds silently**.

For a super admin completing a request from another school's subdomain that is: the sale is recorded against the wrong school (wrong revenue, wrong payout, wrong `school_percentage_snapshot`), and the request is left at `payment_received` forever while the student is already entitled. `confirmPaymentReceived` (line ~397) has the same shape.

## Why it is latent, not live

The detail page (`app/[locale]/dashboard/admin/payment-requests/[requestId]/page.tsx:81`) filters by tenant and redirects, so there is no UI that reaches it — proven by the last case in `tests/playwright/manual-sale-settlement.spec.ts`. It is one route or one `!superAdmin` away from being live.

## Fix

Use `request.tenant_id` for every write in both actions (insert, update filter, notification fan-out), or drop the super-admin read bypass so the two halves cannot disagree. Prefer the first: the bypass is presumably there on purpose for support work.

## Acceptance

- A super admin completing a request that belongs to tenant B while on tenant A's host produces a transaction with `tenant_id = B` and a request at `completed`.
- Covered where the tenant filter is already asserted, in `manual-sale-settlement.spec.ts`.

### LINKED ISSUE #746: Manual sales are written with payment_provider = NULL
From the PayPal + manual audit (PR #742).

## What happens

`transactions.payment_provider` is nullable with **no default**, and `completeAndEnroll` (`app/actions/payment-requests.ts:510-525`) never sets it. So every offline sale — the rail every school starts on — is a NULL-provider row. Asserted, deliberately, in `tests/playwright/manual-sale-settlement.spec.ts` so the assumption is visible rather than folklore.

## Why nothing is broken today

Every reader goes through `resolveProvider()` (`lib/payments/revenue-share.ts`), which coalesces "no provider and no Stripe payment-intent id" to `manual` — the same coalesce `get_platform_revenue` does in SQL. `app/actions/platform/payouts.ts:53` filters `.in('payment_provider', PLATFORM_SETTLED_PROVIDERS)`, which excludes NULL and excludes `manual` alike, so the payouts view is correct by accident.

## Why it should still be fixed

The invariant is "ask the capability, not the identity" (#547), and it is enforced nowhere for this column. Any new reader that does `.eq('payment_provider', 'manual')`, groups by the column, or joins on it silently drops every offline sale — and offline is the majority rail for a new school. The next person to write that query has no way to know.

## Fix

1. `ALTER TABLE public.transactions ALTER COLUMN payment_provider SET DEFAULT 'manual'` plus a backfill of existing NULL rows that carry no `stripe_payment_intent_id` (the ones that do are legacy Stripe — `resolveProvider`'s other branch — and must become `stripe`, not `manual`).
2. Set it explicitly in `completeAndEnroll` anyway, so the insert says what it means.
3. Keep `resolveProvider()`: it still has to answer for pre-backfill history if the backfill is ever partial.

## Acceptance

- A fresh manual completion writes `payment_provider = 'manual'` (flip the assertion in `manual-sale-settlement.spec.ts`).
- The school revenue figures and `getPayoutsOwed` are unchanged before and after the backfill on a seeded database.

### LINKED ISSUE #749: /checkout/manual renders a payment form for a $0 product and only fails on submit
From the PayPal + manual audit (PR #742).

## What the student sees

A free offering is stored as a `manual` product at price 0 (the product wizard's NOT NULL default, #727). `app/[locale]/checkout/manual/page.tsx` does not check the price: it renders the whole "request payment instructions" form, priced at `$0.00`. Only when the student fills it in and presses the button does `createPaymentRequest` throw `This product is free — enroll directly from the product page`, which lands as an error toast on a form that then just sits there.

The guard itself is right and worth keeping — without it every giveaway would mint an unpayable bill in the admin's queue. It is pinned by the second case in `tests/playwright/manual-sale-settlement.spec.ts`. The problem is that it fires after the work, with no way forward.

## Fix

In the product branch of the page (the `parseFloat(product.price)` it already computes), send a $0 product to the one-click enrolment the product page offers instead of rendering the form — the same route `app/[locale]/(public)/checkout/page.tsx` takes for a course with no paid product (`redirect('/courses/<id>?enroll=1')`). Check the plan branch for the same shape.

## Acceptance

- `/checkout/manual?productId=<free product>` never shows a payment form.
- The server guard stays: calling `createPaymentRequest` for a $0 product still throws, and the existing spec case is updated to assert the redirect instead of the toast.

guillermoscript:
## What

Three fixes to the manual (offline) rail from the #742 audit: the sale lands on the request's own school, it is labelled `manual`, and a $0 item never reaches the request form.

Closes #745
Closes #746
Closes #749

> **Stacked on #742** (the manual settlement spec lives there). Until #742 merges this diff also shows its commits; this change is the last commit.

## Why

- **#745** — `completeAndEnroll` / `confirmPaymentReceived` read with a super-admin bypass but wrote `tenant_id` and filtered the update by the *acting* host's tenant. A cross-host completion booked revenue, payout and fee snapshot to the wrong school, and the request update matched nothing yet "succeeded".
- **#746** — manual sales were `payment_provider = NULL`; any reader not going through `resolveProvider()` silently dropped every offline sale.
- **#749** — `/checkout/manual` rendered a full payment form for a $0 product, refused only on submit.

## How

- Actions use `request.tenant_id` for the insert and both update filters, and `.select()` the update so zero rows is an error, not silence.
- `completeAndEnroll` writes `payment_provider: 'manual'`. Migration `20260914140000` sets the column default to `'manual'` and backfills NULLs with `resolveProvider`'s exact rule (Stripe intent → `stripe`, else `manual`) — with `DISABLE TRIGGER USER` around the UPDATE, because `after_transaction_update` would re-run `enroll_user`/`handle_new_subscription` and `before_transaction_split_snapshot_update` would re-stamp legacy snapshots with today's split (#496).
- `/checkout/manual`: $0 product → `/products/<id>` (one-click enroll); $0 plan → `/checkout?planId=` (free subscribe). `createPaymentRequest`'s server guard stays.

## How to QA

1. `npm run db:reset`, then on `default.lvh.me:3000` as `student@e2etest.com`: open `/en/checkout/manual?productId=<a $0 manual product>` → lands on `/en/products/<id>`, no form.
2. `npx playwright test manual-sale-settlement --workers=1` — free case asserts the redirect; completion asserts `payment_provider = 'manual'`.
3. Backfill check: before/after the migration, `select payment_provider, count(*) from transactions group by 1` shows no NULLs after, and `select count(*) from subscriptions` / entitlements counts are unchanged.

**Not covered by E2E:** #745's cross-host super-admin path. No UI reaches it (the detail page filters by tenant and redirects — the last case in the spec), which is why it was latent.

## Screenshots / GIF

No visible change beyond a redirect replacing a form that could not be submitted; covered by the spec assertion.

## Checklist

- [ ] `npm run build` passes (CI)
- [x] Every new tenant-scoped query filters by `tenant_id` (now the request's own)
- [x] Tested with every relevant role (student page; admin actions via spec)
- [x] New UI strings added to both `messages/en.json` and `messages/es.json` (none)
- [ ] Migration applies cleanly on `npm run db:reset` (CI — local Supabase was down), has RLS policies (no new table), `lib/database.types.ts` unchanged (default only)


guillermoscript:
**Migration `20260914140000` verified on local Supabase** (applied via `supabase migration up --local`; trigger state after: all four `transactions` triggers `enabled=O`, column default `'manual'`, 0 NULL providers).

Trigger-suspension proof, in a rolled-back transaction on a legacy-shaped row (successful product sale, `payment_provider` NULL, `school_percentage_snapshot` NULL, entitlements removed):

| Run | entitlements recreated | snapshot |
|---|---|---|
| Migration backfill (triggers suspended) | 0 ✅ | NULL ✅ (provider → `manual`) |
| Control — same label change, triggers ON | 1 ❌ | 80.00 ❌ |

So without `DISABLE TRIGGER USER` the backfill would have re-granted access and stamped today's split onto history (#496). Local data had only one transaction, hence the synthetic row.

PR #752, opened 2026-09-14
feat(#744): PayPal for school → platform billing

### LINKED ISSUE #744: Decide PayPal for school → platform billing: implement it, or close the endpoint it cannot use
From the PayPal + manual audit (PR #742).

## State today

`PROVIDER_CAPABILITIES.paypal.supportsPlatformBillingCheckout` is **false** (`lib/payments/paypal-provider.ts`), so `/api/billing/checkout` refuses PayPal and no school can ever start a PayPal platform checkout. But `PLATFORM_WEBHOOK_PROVIDERS` (`lib/billing/platform-billing.ts:37`) **includes `paypal`**, so `/api/billing/webhook/paypal` is live.

An endpoint with no legitimate producer. Every event it can receive for PayPal today comes, by construction, from the student loop — which is exactly the class of bug PR #742 had to guard against in the dispatcher (`STUDENT_LOOP_EVENT_TYPES`).

`tests/unit/platform-webhook-loop-isolation.test.ts` pins both halves, so whichever way this goes the test has to be updated deliberately.

## Option A — close it

Drop `paypal` from `PLATFORM_WEBHOOK_PROVIDERS`. The endpoint 404s, which is what `solana` and `manual` already do by design. Cheapest, and loses nothing that works.

## Option B — implement it

PayPal has native Billing Plans and renewal webhooks, so the rail is capable:

- flip `supportsPlatformBillingCheckout: true`;
- `platform_plan_prices` rows for PayPal (a Billing Plan id per plan × interval), created the way the other catalog rails are;
- platform checkout goes through Billing Subscriptions with `tenant_id` / `plan_id` / `plan_slug` / `interval` in `custom_id` — note PayPal gives us **one 127-char string**, which already carries three fields for the student loop, so the platform packing needs its own encoding and a length check;
- `supportsPlanChange` stays false → plan changes keep the app-side supersession path.

## Decide on

Whether schools paying the platform by PayPal is a market we want (LATAM overlap with the crypto rails is the real question), or whether Stripe + Lemon Squeezy + Binance + Solana + bank transfer is enough. Answer here, then do A or B in one PR.

guillermoscript:
## What

This PR makes PayPal a working rail for **school → platform billing** (Option B from #744). A school admin can now subscribe to a platform plan through PayPal Billing Subscriptions, and `/api/billing/webhook/paypal` finally has a legitimate producer.

Four gaps surfaced while surveying. Flipping the capability alone would have shipped each of them as a bug, so this PR closes them too, including one that already affects Lemon Squeezy and Binance today.

Closes #744

## Why

`supportsPlatformBillingCheckout` was `false` for PayPal, yet its platform webhook endpoint was live. #744 asked us to either close the endpoint or implement the rail. The decision was to implement it. The changes:

**1. The rail itself**
- PayPal `supportsPlatformBillingCheckout: true`, in both the static table and the adapter class.
- A platform checkout (`hosted: true`) packs its correlation into PayPal's single 127-character `custom_id` with its own tagged encoding:
  - format: `plt|<tenant_id>|<plan_id>|<m|y>|<billing_switch_id?>` (116 characters worst case);
  - throws above 127 instead of letting PayPal reject or truncate it;
  - never shares keys with the student `reference|userId|tenantId` shape.
- `plan_slug` does not fit in the id, so the dispatcher looks it up from `plan_id`. Otherwise `tenants.plan`, which every feature gate reads, would not change.
- The Stripe `{CHECKOUT_SESSION_ID}` template in the shared success URL gets its braces percent-encoded for PayPal's `return_url`.

**2. A second webhook id.** PayPal signs each delivery over the id of the webhook *registration* it was sent to, and every registration on an app receives every event.
- The platform endpoint verifies against a new `PAYPAL_PLATFORM_WEBHOOK_ID`. If it is unset, verification fails closed and never falls back to the student id.
- Platform checkout readiness now requires the variable, so a school is never offered a rail that cannot activate.

**3. Student plan purchases could rewrite a school's platform subscription (every shared-account rail).**
- `STUDENT_LOOP_EVENT_TYPES` only caught `payment.*` / `refund.*` events.
- A student's PayPal or Lemon Squeezy *plan* purchase is a `subscription.activated` event carrying `{userId, tenantId, planId}`. The platform dispatcher accepts both `tenantId` and `planId`, so on a bank-transfer school (no `provider_subscription_id`) the event reached the upsert.
- The dispatcher now ignores events carrying the student checkout's `userId` without the platform checkout's `tenant_id`. No platform checkout has ever set `userId`, including the legacy `/api/stripe/checkout-session` metadata, which was checked.

**4. A PayPal cancel is final and immediate.** Calling cancel sends `BILLING.SUBSCRIPTION.CANCELLED` right away, and the dispatcher used to downgrade to free on it, taking back days the school had paid for. The fix:
- New capability `supportsScheduledCancellation`: true for Stripe and Lemon Squeezy, false elsewhere.
- On a rail without it, a CANCELLED event for the current subscription with a future `current_period_end` keeps the plan and records `cancel_at_period_end`.
- The expiry cron's cancel phase (phase 4 only) now also walks those rails (`PLATFORM_APP_CANCELED_PROVIDERS`), so the plan ends when the period does.
- `subscription.expired` still downgrades immediately.
- The billing screen hides **Reactivate** on such a rail, where the action could only fail, and shows a short hint to choose a plan again (en + es).

**5. There was no way to change plan on PayPal.** The issue assumed plan changes would "keep the app-side supersession path", but for the same rail that path did not exist:
- the upgrade page sends non-plan-change rails to `/api/billing/checkout`;
- that route answered 400 "already have an active subscription";
- `changePlan` refused;
- `platform_subscription_switches` had `CHECK (source_payment_provider <> target_payment_provider)`.

Now:
- The migration `20260914150000_platform_switch_same_rail_supersession.sql` drops that CHECK. The rule moves into `beginPlatformSubscriptionSwitch` via `supersedesOnSameRail()`, which is true only for a native-subscription rail with no in-place plan change and no self-managed period (PayPal today).
- The checkout route allows a same-rail PayPal checkout for a different plan or interval, or after a cancellation. It still refuses the identical live plan.
- On activation, the existing switch machinery promotes the new subscription (clearing `cancel_at_period_end`) and cancels the old one. The old subscription's CANCELLED event is handled by `recordSupersededTerminalEvent`.
- Stripe and Lemon Squeezy keep their in-place change, and self-managed rails keep their renewal path.

**6. Price catalog.** PayPal rows are entered the way the other catalog rails' are: the super admin pastes the Billing Plan id under Platform → Plans. The server rejects anything that is not `P-…`, since a pasted `PROD-…` product id would only fail after a school picked the plan. The editor shows a PayPal-specific placeholder.

**Known trade-offs and limits**
- A same-rail PayPal plan change bills the new plan from activation and cancels the old subscription with no proration. Cross-rail switches already behave this way.
- **Pre-existing, not changed here:** a school whose platform subscription is `past_due` cannot start a switch (`beginPlatformSubscriptionSwitch` requires `active`). A replacement activation then fails the identity guard on any rail.
- **Not exercised against live PayPal credentials.** Everything is covered by unit fakes. The PayPal settlement E2E stub covers the student loop only.

**Deploy (production):**
1. In the PayPal app, register a second webhook → `https://<domain>/api/billing/webhook/paypal`, with events `BILLING.SUBSCRIPTION.ACTIVATED`, `.CANCELLED`, `.EXPIRED`, `.SUSPENDED`, `.PAYMENT.FAILED` and `PAYMENT.SALE.COMPLETED`.
2. Set `PAYPAL_PLATFORM_WEBHOOK_ID` in the Dokploy runtime env (runtime only, not a build arg).
3. Create one PayPal Billing Plan per plan × interval and paste the `P-…` ids under Platform → Plans. Enable PayPal for the tenant (`paypal_enabled`).
4. Apply the migration to cloud.

Until step 2 is done, PayPal shows as "not configured" on the upgrade screen rather than half-working.

## How to QA

On a fresh `npm run db:reset`, with `PAYPAL_CLIENT_ID`/`PAYPAL_CLIENT_SECRET` set to sandbox credentials and `NEXT_PUBLIC_PLATFORM_DOMAIN=lvh.me:3000`:

1. **Readiness gate.** Log in as `owner@e2etest.com` / `password123` (super admin) at `http://lvh.me:3000/en/platform/plans`.
   - Add a PayPal monthly price to **Pro** with id `PROD-123`. Expect the error "A PayPal price must be a Billing Plan id (P-…), not a product id".
   - Save it with a sandbox Billing Plan id `P-…` instead. With `PAYPAL_PLATFORM_WEBHOOK_ID` unset, the row is flagged as missing credentials and PayPal is not offered on `/en/dashboard/admin/billing/upgrade`.
2. **Checkout.** Set `PAYPAL_PLATFORM_WEBHOOK_ID` to your sandbox webhook id (a webhook pointed at `/api/billing/webhook/paypal`, e.g. through a tunnel) and restart. As the Default School admin, choose Pro → PayPal.
   - You are redirected to PayPal's approval page.
   - After approving, the webhook activates the plan: Billing shows Pro, `platform_subscriptions.payment_provider = 'paypal'` with an `I-…` id, and `tenants.plan = 'pro'`.
3. **Plan change on the same rail.** From the upgrade page choose Starter → PayPal and approve. On activation the row moves to Starter with a new `I-…` id, and the old PayPal subscription is cancelled in the sandbox. Choosing the plan you are already on returns "You are already subscribed to this plan on this payment method."
4. **Cancel keeps the paid period.** Click **Cancel plan**.
   - The plan stays active, with "cancels on <period end>" and the new hint that the cancellation cannot be undone. No **Reactivate** button.
   - After the CANCELLED webhook the row is still `active` with `cancel_at_period_end = true`.
   - Move `current_period_end` into the past and run `curl -H "Authorization: Bearer $CRON_SECRET" http://lvh.me:3000/api/cron/expire-platform-subscriptions`. The school is downgraded to free.
5. **Regression: Stripe and Lemon Squeezy unchanged.** A second checkout on the same Stripe/LS rail is still refused with "Change your plan from the billing page instead", and their cancel still shows **Reactivate**.

## Screenshots / GIF

Screenshots follow in a comment.

## Checklist

- [x] `npm run build` passes
- [x] Every new tenant-scoped query filters by `tenant_id` (or the table genuinely has no such column): the dispatcher's final-cancel read/update filters `tenant_id` + provider + subscription id. The `platform_plans` slug lookup is a global table.
- [ ] Tested with every relevant role (student / teacher / admin): admin billing view checked locally, see screenshots. Super-admin price editor validation covered by the server check. No live PayPal sandbox run.
- [x] Loading and error states handled: new refusals return 400 with specific messages. A missing webhook id surfaces as "not configured".
- [x] Migration (if any) applies cleanly (`supabase migration up --local` applied it; the remaining CHECK constraints were verified intact). It adds no table or RLS change. `lib/database.types.ts` is unaffected (a constraint drop only).

### Checks run

- `npm run typecheck`: clean
- `npx eslint` on every changed source file: 0 errors (2 pre-existing `_productId` warnings in `paypal-provider.ts`)
- `npm run build`: passed
- `npm run test:unit`: 101 files, 1256 tests passed. New and updated coverage:
  - `platform-paypal-cancel.test.ts` (new): a final cancel keeps the paid period; downgrades once it has passed; Stripe and `expired` unchanged.
  - `platform-webhook-loop-isolation.test.ts`: a student plan purchase on PayPal, Lemon Squeezy or Binance leaves the school's row untouched; slug lookup; the endpoint invariant now has no PayPal carve-out.
  - `paypal-binance-providers.test.ts`: platform `custom_id` round-trip, the 127-character limit, disjoint keys between loops, hosted checkout packing and brace encoding, `verifyWebhook` failing closed on an empty id.
  - `platform-billing-checkout.test.ts`: a same-rail PayPal plan change begins a switch; the same live plan is refused; re-buying after a cancel is allowed; the new provider lists.
  - `platform-subscription-switch.test.ts`, `expire-platform-subscriptions.test.ts`, `platform-checkout-availability.test.ts`, `provider-capability-parity.test.ts`: same-rail switch rule, phase 4 on PayPal, readiness requiring `PAYPAL_PLATFORM_WEBHOOK_ID`, the new capability's values.
- `supabase migration up --local`: applied. `platform_subscription_switches` keeps its six remaining CHECKs; the `source <> target` one is gone.

https://claude.ai/code/session_019A72XZ6abyj8YLadHhpMLV


guillermoscript:
## Visual verification: billing overview after a PayPal cancellation

These were captured locally with Playwright, logged in as `owner@e2etest.com` on Default School, with a temporary `platform_subscriptions` fixture: Pro, monthly, `active`, `cancel_at_period_end = true`, period ending in 20 days. The fixture was removed afterwards.

**After (PayPal).** The plan stays active until the period end. **Reactivate plan** is gone, because a PayPal cancel is final at PayPal and the action could only fail. The new hint tells the admin to choose a plan again.

After: PayPal subscription cancelled

**Contrast (Stripe, same fixture).** A rail that only *schedules* its cancellation still shows **Reactivate plan** and no hint. This is also what the PayPal row looked like before this PR, with a Reactivate button that would have errored.

Contrast: Stripe subscription cancelled

**Not included.**
- No flow GIF: the browser-automation extension cannot load the local `lvh.me` tenant domain.
- No true "before" capture: implementation started before a before-shot was taken. The Stripe contrast shows the pre-change rendering.
- No PayPal sandbox run of checkout or approval; that is QA steps 2–4 in the PR body.

Please check the two screenshots before merging. The only other UI change is the PayPal placeholder (`P-… (Billing Plan id)`) in the Platform → Plans price editor.

